PR #22, opened 2020-12-09
Do not start container until after attach. Do not attach if container never starts.

innovate-invent:
This deals with a race condition that causes an error when the container starts and exits before the provider can attach.


innovate-invent:
bump

innovate-invent:
bump

mavogel:
are you still on this @innovate-invent ? Maybe wait with the rebase until #145 is merged to avoid too many conflicts

innovate-invent:
Yea, looks like the refactor is going to mess this up. I'll wait.

mavogel:
i think we're ready to continue @innovate-invent  :)

PR #23, opened 2020-12-09
Update to SDK v2

innovate-invent:
Some more work needs to be done to remove logic that deals with timeouts in place of context cancellation.


mavogel:
closed in favour of #102

PR #35, opened 2020-12-22
feat: support Docker plugin

### LINKED ISSUE #24: Add support for docker plugins
Terraform Version

0.13.5
Affected Resource(s)

docker_plugin
Expected Behavior

Configure and manage docker plugins on a daemon
Actual Behavior

Not implemented
References

- https://github.com/hashicorp/terraform-provider-docker/issues/307
- https://github.com/moby/moby/blob/master/api/types/plugin.go
- https://docs.docker.com/engine/reference/commandline/plugin/
- https://github.com/moby/moby/blob/master/client/plugin_install.go

suzuki-shunsuke:
Close #24

### DIFF
--- a/docker/data_source_docker_plugin.go
+++ b/docker/data_source_docker_plugin.go
@@ -46,22 +46,27 @@ func dataSourceDockerPlugin() *schema.Resource {
 	}
 }
 
-func getDataSourcePluginKey(d *schema.ResourceData) string {
-	if id, ok := d.GetOk("id"); ok {
-		return id.(string)
-	}
-	if alias, ok := d.GetOk("alias"); ok {
-		return alias.(string)
-	}
-	return ""
-}
-
 var errDataSourceKeyIsMissing = errors.New("One of id or alias must be assigned")
 
+func getDataSourcePluginKey(d *schema.ResourceData) (string, error) {
+	id, idOK := d.GetOk("id")
+	alias, aliasOK := d.GetOk("alias")
+	if idOK {
+		if aliasOK {
+			return "", errDataSourceKeyIsMissing
+		}
+		return id.(string), nil
+	}
+	if aliasOK {
+		return alias.(string), nil
+	}
+	return "", errDataSourceKeyIsMissing
+}
+
 func dataSourceDockerPluginRead(d *schema.ResourceData, meta interface{}) error {
-	key := getDataSourcePluginKey(d)
-	if key == "" {
-		return errDataSourceKeyIsMissing
+	key, err := getDataSourcePluginKey(d)
+	if err != nil {
+		return err
 	}
 	client := meta.(*ProviderConfig).DockerClient
 	ctx := context.Background()
--- a/docker/data_source_docker_plugin.go
+++ b/docker/data_source_docker_plugin.go
@@ -28,7 +28,7 @@ func dataSourceDockerPlugin() *schema.Resource {
 				Description: "Docker Plugin Reference.",
 				Optional:    true,
 			},
-			"disabled": {
+			"enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
--- a/docker/data_source_docker_plugin.go
+++ b/docker/data_source_docker_plugin.go
@@ -20,12 +20,12 @@ func dataSourceDockerPlugin() *schema.Resource {
 			"alias": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Description: "Docker Plugin alias.",
+				Description: "Docker Plugin alias",
 			},
 
 			"plugin_reference": {
 				Type:        schema.TypeString,
-				Description: "Docker Plugin Reference.",
+				Description: "Docker Plugin Reference",
 				Computed:    true,
 			},
 			"enabled": {
--- a/docker/data_source_docker_plugin.go
+++ b/docker/data_source_docker_plugin.go
@@ -37,7 +37,7 @@ func dataSourceDockerPlugin() *schema.Resource {
 				Computed:    true,
 				Description: "If true, grant all permissions necessary to run the plugin",
 			},
-			"args": {
+			"env": {
 				Type:     schema.TypeSet,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
--- a/docker/data_source_docker_plugin.go
+++ b/docker/data_source_docker_plugin.go
@@ -1,0 +1,75 @@
+package docker
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceDockerPlugin() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDockerPluginRead,
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"alias": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Docker Plugin alias.",
+			},
+
+			"plugin_reference": {
+				Type:        schema.TypeString,
+				Description: "Docker Plugin Reference.",
+				Optional:    true,
+			},
+			"disabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"grant_all_permissions": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, grant all permissions necessary to run the plugin",
+			},
+			"args": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func getDataSourcePluginKey(d *schema.ResourceData) string {
+	if id, ok := d.GetOk("id"); ok {
+		return id.(string)
+	}
+	if alias, ok := d.GetOk("alias"); ok {
+		return alias.(string)
+	}
+	return ""
+}
+
+var errDataSourceKeyIsMissing = errors.New("One of id or alias must be assigned")
+
+func dataSourceDockerPluginRead(d *schema.ResourceData, meta interface{}) error {
+	key := getDataSourcePluginKey(d)
+	if key == "" {
+		return errDataSourceKeyIsMissing
+	}
+	client := meta.(*ProviderConfig).DockerClient
+	ctx := context.Background()
+	plugin, _, err := client.PluginInspectWithRaw(ctx, key)
+	if err != nil {
+		return fmt.Errorf("inspect a Docker plugin "+key+": %w", err)
+	}
+
+	setDockerPlugin(d, plugin)
+	return nil
+}
--- a/docker/data_source_docker_plugin.go
+++ b/docker/data_source_docker_plugin.go
@@ -26,20 +26,20 @@ func dataSourceDockerPlugin() *schema.Resource {
 			"plugin_reference": {
 				Type:        schema.TypeString,
 				Description: "Docker Plugin Reference.",
-				Optional:    true,
+				Computed:    true,
 			},
 			"enabled": {
 				Type:     schema.TypeBool,
-				Optional: true,
+				Computed: true,
 			},
 			"grant_all_permissions": {
 				Type:        schema.TypeBool,
-				Optional:    true,
+				Computed:    true,
 				Description: "If true, grant all permissions necessary to run the plugin",
 			},
 			"args": {
 				Type:     schema.TypeSet,
-				Optional: true,
+				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 		},
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -24,6 +24,7 @@ func resourceDockerPlugin() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 				Optional:    true,
+				ForceNew:    true,
 				Description: "Docker Plugin alias.",
 			},
 			"disabled": {
@@ -35,6 +36,12 @@ func resourceDockerPlugin() *schema.Resource {
 				Optional:    true,
 				Description: "If true, grant all permissions necessary to run the plugin",
 			},
+			"env": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
 			"disable_when_set": {
 				Type:        schema.TypeBool,
 				Optional:    true,
@@ -43,11 +50,6 @@ func resourceDockerPlugin() *schema.Resource {
 			"force_destroy": {
 				Type:     schema.TypeBool,
 				Optional: true,
-			},
-			"env": {
-				Type:     schema.TypeSet,
-				Optional: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"enable_timeout": {
 				Type:        schema.TypeInt,
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -16,7 +16,7 @@ func resourceDockerPlugin() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"plugin_reference": {
 				Type:        schema.TypeString,
-				Description: "Docker Plugin Reference.",
+				Description: "Docker Plugin Reference",
 				Required:    true,
 				ForceNew:    true,
 			},
@@ -25,7 +25,7 @@ func resourceDockerPlugin() *schema.Resource {
 				Computed:    true,
 				Optional:    true,
 				ForceNew:    true,
-				Description: "Docker Plugin alias.",
+				Description: "Docker Plugin alias",
 			},
 			"enabled": {
 				Type:     schema.TypeBool,
@@ -78,7 +78,7 @@ func resourceDockerPlugin() *schema.Resource {
 			"force_disable": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "If true, then the plugin is disabled forcibly when the plugin is disabled.",
+				Description: "If true, then the plugin is disabled forcibly when the plugin is disabled",
 			},
 		},
 	}
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -14,27 +14,50 @@ func resourceDockerPlugin() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 		Schema: map[string]*schema.Schema{
-			"name": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+			"plugin_reference": {
+				Type:        schema.TypeString,
+				Description: "Docker Plugin Reference.",
+				Required:    true,
+				ForceNew:    true,
+			},
+			"alias": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Optional:    true,
+				Description: "Docker Plugin alias.",
 			},
 			"disabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
-			"destroy_option": {
-				Type:     schema.TypeList,
+			"grant_all_permissions": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, grant all permissions necessary to run the plugin",
+			},
+			"disable_when_set": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, the plugin becomes disabled temporarily when the plugin setting is updated",
+			},
+			"force_destroy": {
+				Type:     schema.TypeBool,
 				Optional: true,
-				MaxItems: 1,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"force": {
-							Type:     schema.TypeBool,
-							Optional: true,
-						},
-					},
-				},
+			},
+			"env": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"enable_timeout": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "HTTP client timeout to enable the plugin",
+			},
+			"force_disable": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, then the plugin is disabled forcely when the plugin is disabled.",
 			},
 		},
 	}
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -66,11 +66,6 @@ func resourceDockerPlugin() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 
-			"disable_when_set": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				Description: "If true, the plugin becomes disabled temporarily when the plugin setting is updated",
-			},
 			"force_destroy": {
 				Type:     schema.TypeBool,
 				Optional: true,
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -37,7 +37,7 @@ func resourceDockerPlugin() *schema.Resource {
 				Optional:    true,
 				Description: "If true, grant all permissions necessary to run the plugin",
 			},
-			"args": {
+			"env": {
 				Type:     schema.TypeSet,
 				Optional: true,
 				Computed: true,
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -1,0 +1,28 @@
+package docker
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceDockerPlugin() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceDockerPluginCreate,
+		Read:   resourceDockerPluginRead,
+		Update: resourceDockerPluginUpdate,
+		Delete: resourceDockerPluginDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"disabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+		},
+	}
+}
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -42,9 +42,7 @@ func resourceDockerPlugin() *schema.Resource {
 				Type:          schema.TypeSet,
 				Optional:      true,
 				ConflictsWith: []string{"grant_all_permissions"},
-				Set: func(v interface{}) int {
-					return schema.HashString(v.(map[string]interface{})["name"].(string))
-				},
+				Set:           dockerPluginGrantPermissionsSetFunc,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -23,6 +23,19 @@ func resourceDockerPlugin() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"destroy_option": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"force": {
+							Type:     schema.TypeBool,
+							Optional: true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
--- a/docker/resource_docker_plugin.go
+++ b/docker/resource_docker_plugin.go
@@ -14,11 +14,13 @@ func resourceDockerPlugin() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 		Schema: map[string]*schema.Schema{
-			"plugin_reference": {
-				Type:        schema.TypeString,
-				Description: "Docker Plugin Reference",
-				Required:    true,
-				ForceNew:    true,
+			"name": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				Description:      "Docker Plugin name",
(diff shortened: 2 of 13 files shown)


docker/resource_docker_plugin_funcs.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceDockerPluginCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderConfig).DockerClient
+	ctx := context.Background()
+	pluginName := d.Get("name").(string)
+	// TODO
+	if err := client.PluginCreate(ctx, nil, types.PluginCreateOptions{
```
suzuki-shunsuke, line 16: https://pkg.go.dev/github.com/docker/docker/client#Client.PluginCreate

I don't know the content of `createContext`.

docker/resource_docker_plugin.go
```diff
@@ -0,0 +1,24 @@
...
+		Delete: resourceDockerPluginDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
suzuki-shunsuke, line 21: Which attributes are needed?

docker/resource_docker_plugin_funcs.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func resourceDockerPluginUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*ProviderConfig).DockerClient
+	ctx := context.Background()
+	pluginName := d.Get("name").(string)
+	// TODO
+	if err := client.PluginSet(ctx, pluginName, nil); err != nil {
+		return fmt.Errorf("modifiy settings for the Docker plugin "+pluginName+": %w", err)
+	}
```
suzuki-shunsuke, line 46: * PluginUpgrade
* PluginEnable
* PluginDisable

docker/resource_docker_plugin_funcs.go
suzuki-shunsuke: * PluginInstall
* PluginCreate

suzuki-shunsuke:
I have implemented the basic feature.

## Example

```hcl
resource "docker_plugin" "sample-volume-plugin" {
  plugin_reference      = "docker.io/tiborvass/sample-volume-plugin:latest"
  alias                 = "sample-volume-plugin:latest"
  disabled              = true
  grant_all_permissions = true
  disable_when_set      = true
  force_destroy         = true
  enable_timeout        = 60
  force_disable         = true
  env = [
    "DEBUG=1"
  ]
}
```

## Check List

## Check List

- [x] Import
- [x] update
- [ ] install option https://docs.docker.com/engine/reference/commandline/plugin_install/
  - [x] alias
  - [x] disable
  - [ ] disable-content-trust
  - [x] grant-all-permissions
  - [x] key=value
- [x] enable option https://docs.docker.com/engine/reference/commandline/plugin_enable/
  - [x] timeout
- [x] disable option
  - [x] force
- [x] rm option https://docs.docker.com/engine/reference/commandline/plugin_rm/
  - [x] force
- [ ] upgrade plugin

docker/resource_docker_plugin.go
```diff
@@ -0,0 +1,67 @@
...
+				Description: "If true, grant all permissions necessary to run the plugin",
+			},
+			"env": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+
+			"disable_when_set": {
```
suzuki-shunsuke, line 46: I want a better attribute name.

suzuki-shunsuke: To run `docker plugin set`, the plugin must be disabled.

ex.

```
$ docker plugin set 27784976e147 DEBUG=2
Error response from daemon: cannot set on an active plugin, disable plugin before setting
```

docker/resource_docker_plugin.go
```diff
@@ -0,0 +1,67 @@
...
+			"disabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
+			"grant_all_permissions": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, grant all permissions necessary to run the plugin",
+			},
+			"env": {
```
suzuki-shunsuke, line 39: TODO: validate env format. The element must be `key=value`.

suzuki-shunsuke: TODO: rename `env` to `args`.
https://docs.docker.com/engine/reference/commandline/plugin_set/#extended-description

suzuki-shunsuke: Oh `ValidateFunc is not yet supported on lists or sets`

```
1 error occurred:
	* resource docker_plugin: args: ValidateFunc is not yet supported on lists or sets.
```

docker/resource_docker_plugin_funcs.go
```diff
@@ -0,0 +1,156 @@
...
+func setDockerPlugin(d *schema.ResourceData, plugin *types.Plugin) {
+	d.SetId(plugin.ID)
+	d.Set("plugin_reference", plugin.PluginReference)
+	d.Set("alias", plugin.Name)
+	d.Set("disabled", !plugin.Enabled)
+	// TODO support other settings
+	// https://docs.docker.com/engine/reference/commandline/plugin_set/#extended-description
+	// source of mounts .Settings.Mounts
+	// path of devices .Settings.Devices
+	// args .Settings.Args
```
suzuki-shunsuke, line 62: I'm not familiar with these features, so I don't implement them in this pull request.

suzuki-shunsuke:
@innovate-invent 
I have implemented the `docker_plugin` resource. #24 
document: https://github.com/kreuzwerker/terraform-provider-docker/blob/feat/support-docker-plugin/website/docs/r/plugin.html.markdown

Would you check this?

docker/resource_docker_plugin.go
innovate-invent: what is the behavior when disable_when_set is false?

docker/resource_docker_plugin_funcs.go
innovate-invent: this is implemented by setting all attributes that are not of that list to ForceNew: true

docker/resource_docker_plugin.go
```diff
@@ -0,0 +1,67 @@
...
+			},
+			"enable_timeout": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "HTTP client timeout to enable the plugin",
+			},
+			"force_disable": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, then the plugin is disabled forcely when the plugin is disabled.",
```
innovate-invent, line 63: s/forcely/forcibly/

docker/resource_docker_plugin.go
```diff
@@ -0,0 +1,67 @@
...
+	return &schema.Resource{
+		Create: resourceDockerPluginCreate,
+		Read:   resourceDockerPluginRead,
+		Update: resourceDockerPluginUpdate,
+		Delete: resourceDockerPluginDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"plugin_reference": {
```
innovate-invent, line 17: keep consistent with the other resources and have this attribute as `name` rather than `plugin_reference`

docker/resource_docker_plugin_funcs.go
```diff
@@ -0,0 +1,165 @@
...
+	ctx := context.Background()
+	pluginRef := d.Get("plugin_reference").(string)
+	alias := d.Get("alias").(string)
+	log.Printf("[DEBUG] Install a Docker plugin " + pluginRef)
+	body, err := client.PluginInstall(ctx, alias, types.PluginInstallOptions{
+		RemoteRef:            pluginRef,
+		AcceptAllPermissions: d.Get("grant_all_permissions").(bool),
+		Disabled:             d.Get("disabled").(bool),
+		Args:                 getDockerPluginArgs(d.Get("args")),
+	})
```
innovate-invent, line 37: You should implement AcceptPermissionsFunc that matches against permissions listed in a resource attribute that ConfictsWith "grant_all_permissions"


https://pkg.go.dev/github.com/docker/docker/api/types#PluginInstallOptions

docker/resource_docker_plugin_funcs.go
```diff
@@ -0,0 +1,165 @@
...
+			skipArgs = true
+			log.Printf("[DEBUG] Enable a Docker plugin " + pluginID)
+			if err := client.PluginEnable(ctx, pluginID, types.PluginEnableOptions{
+				Timeout: d.Get("enable_timeout").(int),
+			}); err != nil {
+				return fmt.Errorf("enable the Docker plugin "+pluginID+": %w", err)
+			}
+		}
+	}
+	if !skipArgs {
```
innovate-invent, line 140: Can you replace `skipArgs` with ```d.HasChange("disabled") && !d.Get("disabled").(bool)``` to make it more clear what condition the code is executed under?

docker/resource_docker_plugin.go
innovate-invent: I just had a look, you should remove this attribute and just disable the plugin on all updates.

docker/resource_docker_plugin_funcs.go
```diff
@@ -0,0 +1,165 @@
...
+	d.SetId(plugin.ID)
+	d.Set("plugin_reference", plugin.PluginReference)
+	d.Set("alias", plugin.Name)
+	d.Set("disabled", !plugin.Enabled)
+	// TODO support other settings
+	// https://docs.docker.com/engine/reference/commandline/plugin_set/#extended-description
+	// source of mounts .Settings.Mounts
+	// path of devices .Settings.Devices
+	// args .Settings.Args
+	d.Set("args", plugin.Settings.Env)
```
innovate-invent, line 64: Env is not the correct value for Args.
You need to implement all of https://pkg.go.dev/github.com/docker/docker/api/types#PluginSettings

innovate-invent:
Do you mind adding the accompanying data resources as well? They are critical to manage preinstalled plugins.

docker/resource_docker_plugin.go
suzuki-shunsuke: When we try to update a active plugin

1. If disable_when_set is false (default), the update would fail
1. If disable_when_set is true, the plugin becomes disable temporarily

I introduce this attribute because I concern someone may prefer the failure rather than temporal disable.

But I think removing this attribute is possibility too because it is simpler.

innovate-invent: Terraforms workflow shouldn't fail when updating a resource. There would be no way in terraform to explicitly disable, update, and reenable. I think it would be best to remove this attribute.

docker/resource_docker_plugin.go
suzuki-shunsuke: Fixed 7a79c28c88fd51de0cdb2edd843c63da55ab047e

docker/resource_docker_plugin_funcs.go
suzuki-shunsuke: https://github.com/kreuzwerker/terraform-provider-docker/blob/7a79c28c88fd51de0cdb2edd843c63da55ab047e/docker/resource_docker_plugin_funcs.go#L131

skipArgs is used to avoid the unneeded changes.
But the current code is a little complicated and we may be able to refactor.

suzuki-shunsuke:
> Do you mind adding the accompanying data resources as well? They are critical to manage preinstalled plugins.

I'll try it.

suzuki-shunsuke:
Thank you for your reivew.

docker/resource_docker_plugin_funcs.go
innovate-invent: This document may help: https://docs.docker.com/engine/extend/config/

innovate-invent: I think I misunderstood the purpose of those settings. Those would be overrides for settings that are already configured when the plugin was created.

docker/resource_docker_plugin_funcs.go
innovate-invent: skipArgs == d.HasChange("disabled") && !d.Get("disabled").(bool) when it is evaluated.

docker/resource_docker_plugin_funcs.go
innovate-invent: I misunderstood the purpose of this value. You might want to change the attribute to `env` to make clear what it is doing.

docker/data_source_docker_plugin.go
```diff
@@ -0,0 +1,75 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"alias": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Docker Plugin alias.",
+			},
+
+			"plugin_reference": {
```
innovate-invent, line 26: all but id and alias need to be Computed: true.
id and alias should conflict

innovate-invent:
Can you also add a data source that takes no attributes calls https://pkg.go.dev/github.com/docker/docker@v1.13.1/client#Client.PluginList 
 and returns a list of { id, name, alias, enabled } maps

docker/data_source_docker_plugin.go
```diff
@@ -0,0 +1,75 @@
...
+				Type:        schema.TypeString,
+				Description: "Docker Plugin Reference.",
+				Optional:    true,
+			},
+			"disabled": {
```
innovate-invent, line 31: do you think this should be changed to 'enabled'?
A default should also be set for this attribute.

docker/data_source_docker_plugin.go
```diff
@@ -0,0 +1,75 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"alias": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Docker Plugin alias.",
```
innovate-invent, line 23: You might want to remove all '.' from the end of the descriptions

docker/data_source_docker_plugin.go
```diff
@@ -46,22 +46,27 @@ func dataSourceDockerPlugin() *schema.Resource {
 	}
 }
 
-func getDataSourcePluginKey(d *schema.ResourceData) string {
-	if id, ok := d.GetOk("id"); ok {
-		return id.(string)
+var errDataSourceKeyIsMissing = errors.New("One of id or alias must be assigned")
```
innovate-invent, line 49: You should replace this with the [ExactlyOneOf](https://github.com/hashicorp/terraform-plugin-sdk/blob/aa07c4354f9a8fd948ed40f020f340132ada44bc/helper/schema/schema.go#L196) attribute property

innovate-invent:
You are going to want to rebase on https://github.com/kreuzwerker/terraform-provider-docker/pull/102 as it will break your PR

docker/resource_docker_plugin.go
```diff
@@ -0,0 +1,67 @@
...
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "HTTP client timeout to enable the plugin",
+			},
+			"force_disable": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "If true, then the plugin is disabled forcibly when the plugin is disabled.",
+			},
+		},
```
innovate-invent, line 65: Can you add config and settings computed properties that expose the values in the Plugin data.
You can use github.com/mitchellh/mapstructure to automatically convert the go classes to maps before calling d.Set().
See https://github.com/brinkmanlab/terraform-provider-galaxy/blob/09ce5c679ae4625e240fa98e9a46322a645fae1f/galaxy/resource_job.go#L194-L200 for an example.

docker/data_source_docker_plugin.go
suzuki-shunsuke:
```
❯ go vet ./...
# github.com/terraform-providers/terraform-provider-docker/docker
docker/data_source_docker_plugin.go:19:5: unknown field 'ExactlyOneOf' in struct literal of type schema.Schema
docker/data_source_docker_plugin.go:24:5: unknown field 'ExactlyOneOf' in struct literal of type schema.Schema
# github.com/terraform-providers/terraform-provider-docker/docker
vet: docker/data_source_docker_plugin.go:19:5: unknown field ExactlyOneOf in struct literal
```

Oh, `ExactlyOneOf` isn't supported.
Probably we have to update the Docker SDK.

innovate-invent: yes, see https://github.com/kreuzwerker/terraform-provider-docker/pull/102

docker/data_source_docker_plugin.go
suzuki-shunsuke: I want to know what is a problem if `Computed: true` isn't set in Data Sources.

https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed

innovate-invent: https://github.com/hashicorp/terraform-plugin-sdk/blob/aa07c4354f9a8fd948ed40f020f340132ada44bc/helper/schema/schema.go#L78-L79

docker/data_source_docker_plugin.go
suzuki-shunsuke: Fixed. 5f925d659961469a11fa1c16c474475ad15f17a9

docker/resource_docker_plugin_funcs.go
suzuki-shunsuke: Oh, I see.
Fixed. 3ec71c5bc968af149cd565aba0e145d14234eaa7

docker/resource_docker_plugin.go
suzuki-shunsuke: I concern `name` is misunderstood as `alias`.

ex.

```
❯ docker plugin ls
ID             NAME                          DESCRIPTION                         ENABLED
2c6c299f4899   sample-volume-plugin:latest   A sample volume plugin for Docker   false
```

innovate-invent: would 'image' be better?

suzuki-shunsuke:
https://github.com/kreuzwerker/terraform-provider-docker/pull/35#issuecomment-752266786

> Can you also add a data source that takes no attributes and calls https://pkg.go.dev/github.com/docker/docker@v1.13.1/client#Client.PluginList
> and returns a list of { id, name, alias, enabled } maps

Would you explain why such a data source is needed?

innovate-invent:
To detect all installed plugins. The ID can then be used to create a `data "docker_plugin"` data source for select items in the list.

suzuki-shunsuke:
I see. Thank you for your explanation.

suzuki-shunsuke:
This pull request is already large.
So I don't want to add feature in this pull request anymore as much as possible.

docker/data_source_docker_plugin.go
suzuki-shunsuke: Oh, I misunderstood.
I'll fix. Thanks.

suzuki-shunsuke: Fixed. c8b9c7fc8c21c4a40def5e3c0d21ede4b0d87db3

suzuki-shunsuke:
I have created some issues for future work.

* #106
* #107
* #108
* #109
* #110

mavogel:
I propose that we will first merge this PR here and then rebase #102 for the TF SDK update on it. We are ready4review right @suzuki-shunsuke ?

suzuki-shunsuke:
@mavogel Yes. Please review.

docker/resource_docker_plugin.go
suzuki-shunsuke: This attribute is like a force option.
Removing this attribute is like forcibly update of the plugin always.
I wonder if that is better or not.
I'm not familiar with Docker plugin, so I don't know whether it has no problem to disable the plugin temporarily.

@mavogel What do you think?

docker/resource_docker_plugin.go
suzuki-shunsuke: Might be so.
@mavogel Do you have any opinion?

docker/resource_docker_plugin_funcs.go
innovate-invent: I don't know why this comment is being hidden but it is vital to the initial implementation. For production use grant_all_permissions must be false and the permissions need to be explicitly provided.

innovate-invent:
For some reason a critical comment is being hidden by github: https://github.com/kreuzwerker/terraform-provider-docker/pull/35#discussion_r549843380

docker/resource_docker_plugin_funcs.go
mavogel: @suzuki-shunsuke can you take a look, please

suzuki-shunsuke: I have added an attribute `grant_permissions`.

58fccc62e14bd33702e696971771eeb5f518bcd6

TODO

* test
* document

docker/resource_docker_plugin.go
```diff
@@ -33,9 +33,33 @@ func resourceDockerPlugin() *schema.Resource {
...
-				Type:        schema.TypeBool,
-				Optional:    true,
-				Description: "If true, grant all permissions necessary to run the plugin",
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Description:   "If true, grant all permissions necessary to run the plugin",
+				ConflictsWith: []string{"grant_permissions"},
+			},
+			"grant_permissions": {
+				Type:          schema.TypeSet,
```
innovate-invent, line 42: Should this be a TypeMap? keyed on the name

suzuki-shunsuke: TypeMap supports only primitive type.

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema#Schema

> For a TypeMap, it must be a *Schema with a Type that is one of the primitives: TypeString, TypeBool, TypeInt, or TypeFloat

innovate-invent: Ah, right. That is unfortunate.

suzuki-shunsuke:
I'm writing tests of `getDockerPluginGrantPermissions` and found that it doesn't work as expected.
I'm investigating the issue.

suzuki-shunsuke:
https://github.com/kreuzwerker/terraform-provider-docker/pull/35#issuecomment-752807983

> I'm writing tests of `getDockerPluginGrantPermissions` and found that it doesn't work as expected.
> I'm investigating the issue.

Oh, sorry. There is a typo in tests.
It has been solved.

docker/resource_docker_plugin.go
mavogel: I suggest using `name` to be also consistent with the json output of `docker plugin inspect`

docker/resource_docker_plugin.go
mavogel: After reading the docs, I'd also suggest removing the attribute. As a workflow, folks you have to explicitly disable (enable=false) a plugin, update to a new version and then re-enable. If we see later that it's better/desired to use with e.g. `auto_update=true`, we could implement this so the workflow of disable, update and re-enable is done in a function. IMHO.

docker/resource_docker_plugin_funcs.go
mavogel: this conversation can be closed then @suzuki-shunsuke ?

mavogel (changes requested):
See conversations for details. Awesome work till here. Let's get this ready

docker/resource_docker_plugin_funcs.go
suzuki-shunsuke: I think so.
https://github.com/kreuzwerker/terraform-provider-docker/pull/35#issuecomment-752332860

docker/resource_docker_plugin.go
suzuki-shunsuke: Thank you for your reivew.
I see.
Updated. abec1d568c9dae0066d4a7268e74747c999c6c8f 17d8fc6651c17640a7e259885833ecc565678682

suzuki-shunsuke:
Updated. Please review.
https://github.com/kreuzwerker/terraform-provider-docker/pull/35#discussion_r551662238

docker/data_source_docker_plugin.go
mavogel: makes sense @suzuki-shunsuke  to keep in line with the other descriptions.

docker/data_source_docker_plugin.go
mavogel: ok let's merge this PR first then #102 and we can the `ExactlyOneOf` property then

docker/resource_docker_plugin.go
mavogel: @suzuki-shunsuke I think this is also still open

docker/data_source_docker_plugin.go
mavogel: it makes sense @suzuki-shunsuke to keep in line with the other descriptions.

mavogel (changes requested):

docker/resource_docker_plugin.go
mavogel: nice pointer with `mapstructure` @innovate-invent. which properties exactly do you mean to convert?

innovate-invent: Nothing specific, it just provides information about the plugin that are likely important to other resources. A good example is accessing the preconfigured mounts the plugin uses so that a resource that uses it can also mount the mount point.

docker/resource_docker_plugin.go
suzuki-shunsuke: @mavogel 

> I suggest using name to be also consistent with the json output of docker plugin inspect

`name` of `docker plugin inspect` is different from `plugin_reference`.

```
$ docker plugin install --alias foo tiborvass/sample-volume-plugin
$ docker plugin inspect foo | jq -r ".[0].PluginReference"                                                                                                                                                      
docker.io/tiborvass/sample-volume-plugin:latest
$ docker plugin inspect foo | jq -r ".[0].Name"                                                                                                                                                                 
foo:latest
```

```
$ docker version                                                                                                                                                                                                
Client: Docker Engine - Community                                                                                                                                                                               
 Cloud integration: 1.0.4                                                                                                                                                                                       
 Version:           20.10.2                                                                                                                                                                                     
 API version:       1.41                                                                                                                                                                                        
 Go version:        go1.13.15                                                                                                                                                                                   
 Git commit:        2291f61                                                                                                                                                                                     
 Built:             Mon Dec 28 16:12:42 2020                                                                                                                                                                    
 OS/Arch:           darwin/amd64                                                                                                                                                                                
 Context:           default                                                                                                                                                                                     
 Experimental:      true                                                                                                                                                                                        
                                                                                                                                                                                                                
Server: Docker Engine - Community                                                                                                                                                                               
 Engine:                                                                                                                                                                                                        
  Version:          20.10.2                                                                                                                                                                                     
  API version:      1.41 (minimum version 1.12)                                                                                                                                                                 
  Go version:       go1.13.15                                                                                                                                                                                   
  Git commit:       8891c58                                                                                                                                                                                     
  Built:            Mon Dec 28 16:15:23 2020                                                                                                                                                                    
  OS/Arch:          linux/amd64                                                                                                                                                                                 
  Experimental:     true                                                                                                                                                                                        
 containerd:                                                                                                                                                                                                    
  Version:          v1.4.3                                                                                                                                                                                      
  GitCommit:        269548fa27e0089a8b8278fc4fc781d7f65a939b                                                                                                                                                    
 runc:                                                                                                                                                                                                          
  Version:          1.0.0-rc92                                                                                                                                                                                  
  GitCommit:        ff819c7e9184c13b7c2607fe6c30ae19403a7aff                                                                                                                                                    
 docker-init:                                                                                                                                                                                                   
  Version:          0.19.0                                                                                                                                                                                      
  GitCommit:        de40ad0
```

```json
[
    {
        "Config": {
            "Args": {
                "Description": "",
                "Name": "",
                "Settable": null,
                "Value": null
            },
            "Description": "A sample volume plugin for Docker",
            "Documentation": "https://docs.docker.com/engine/extend/plugins/",
            "Entrypoint": [
                "/usr/bin/sample-volume-plugin",
                "/data"
            ],
            "Env": [
                {
                    "Description": "",
                    "Name": "DEBUG",
                    "Settable": [
                        "value"
                    ],
                    "Value": "0"
                }
            ],
            "Interface": {
                "Socket": "plugin.sock",
                "Types": [
                    "docker.volumedriver/1.0"
                ]
            },
            "IpcHost": false,
            "Linux": {
                "AllowAllDevices": false,
                "Capabilities": null,
                "Devices": null
            },
            "Mounts": null,
            "Network": {
                "Type": ""
            },
            "PidHost": false,
            "PropagatedMount": "/data",
            "User": {},
            "WorkDir": "",
            "rootfs": {
                "diff_ids": [
                    "sha256:be0f976063f493d2d3a36e831e62aecbae9ecbaf142eca24a274914b79b96d4a"
                ],
                "type": "layers"
            }
        },
        "Enabled": true,
        "Id": "15868b05e7dc5cd20eb9264b0c781f1f9ba8258e65f3976f703977e05f8814cf",
        "Name": "foo:latest",
        "PluginReference": "docker.io/tiborvass/sample-volume-plugin:latest",
        "Settings": {
            "Args": [],
            "Devices": [],
            "Env": [
                "DEBUG=0"
            ],
            "Mounts": []
        }
    }
]
```

innovate-invent: I don't think I would worry about the idiosyncrasies of the API. The other resources use name to refer to the image to fetch and as a UX point it would be good to be consistent.

docker/data_source_docker_plugin.go
suzuki-shunsuke: Addressed. e3384e9401656632da53dfcd42b1f06c3562847b

docker/resource_docker_plugin.go
suzuki-shunsuke: I added some changes.

7a7aee3025ac7893b3b81638559cc669da077e29
0950bba77c34478329f2c5f34c3a89913c1e3356

* add `DiffSuppressFunc` to alias to complement the latest tag
* add a required field `name`
  * normalize `name` with [reference.ParseAnyReference](https://pkg.go.dev/github.com/docker/distribution/reference#ParseAnyReference) and complement the latest tag
* make `plugin_reference` read only

docker/resource_docker_plugin.go
mavogel: Ok, I'd keep this improvement for the next iteration/a separate issue.

mavogel (approved):
LGTM. Thanks, @suzuki-shunsuke  and @innovate-invent for the fruitful discussion 🚀

PR #98, opened 2020-12-26
feat: add ability to lint/check of links in documentation locally

mavogel:
- implements #40 
- adds linters for the docs and also formatting for the consisting `hcl` code
- adds local make target to have a better/faster workflow, and also the ability for auto-fixing
- also added an additional/enhanced website linter for spellcheck and formatting (so we have 2 atm)

PR #99, opened 2020-12-26
feat: add local semantic commit validation

mavogel:
Currently, the commit messages will be validated locally. #98 should be merged first because it has also a `setup` target


suzuki-shunsuke:
https://github.com/hazcod/semantic-commit-hook

suzuki-shunsuke (approved):

PR #124, opened 2021-01-17
docker container added properties -it (tty and stdin_opn), fixed #120

### LINKED ISSUE #120: Ability to set stdin_open: true and tty: true to a container
Currently is not available to set up the below parameters to docker_container

stdin_open = bool
tty = bool

tanis80:
Added two extra properties to docker container:

tty = true
stdin_open = true

The ability to run: 

docker run -it 

Unit Test added. 
Terraform Provider tested on VPN container with tty and stdin_open required

### DIFF
--- a/docker/resource_docker_container_funcs.go
+++ b/docker/resource_docker_container_funcs.go
@@ -43,17 +43,15 @@ func resourceDockerContainerCreate(d *schema.ResourceData, meta interface{}) err
 		Image:      image,
 		Hostname:   d.Get("hostname").(string),
 		Domainname: d.Get("domainname").(string),
-		Tty:        d.Get("tty").(bool),
-		OpenStdin:  d.Get("stdin_open").(bool),
 	}
 
-	// if value, ok := d["tty"]; ok {
-	// 	config.Tty = value.(bool)
-	// }
+	if value, ok := d.GetOk("tty"); ok {
+		config.Tty = value.(bool)
+	}
 
-	// if value, ok := d["stdin_open"]; ok {
-	// 	config.OpenStdin = d.Get("stdin_open").(bool)
-	// }
+	if value, ok := d.GetOk("stdin_open"); ok {
+		config.OpenStdin = value.(bool)
+	}
 
 	if v, ok := d.GetOk("env"); ok {
 		config.Env = stringSetToStringSlice(v.(*schema.Set))
--- a/docker/resource_docker_container_funcs.go
+++ b/docker/resource_docker_container_funcs.go
@@ -43,14 +43,8 @@ func resourceDockerContainerCreate(d *schema.ResourceData, meta interface{}) err
 		Image:      image,
 		Hostname:   d.Get("hostname").(string),
 		Domainname: d.Get("domainname").(string),
-	}
-
-	if value, ok := d.GetOk("tty"); ok {
-		config.Tty = value.(bool)
-	}
-
-	if value, ok := d.GetOk("stdin_open"); ok {
-		config.OpenStdin = value.(bool)
+		Tty:        d.Get("tty").(bool),
+		OpenStdin:  d.Get("stdin_open").(bool),
 	}
 
 	if v, ok := d.GetOk("env"); ok {
--- a/docker/resource_docker_container_funcs.go
+++ b/docker/resource_docker_container_funcs.go
@@ -43,7 +43,17 @@ func resourceDockerContainerCreate(d *schema.ResourceData, meta interface{}) err
 		Image:      image,
 		Hostname:   d.Get("hostname").(string),
 		Domainname: d.Get("domainname").(string),
+		Tty:        d.Get("tty").(bool),
+		OpenStdin:  d.Get("stdin_open").(bool),
 	}
+
+	// if value, ok := d["tty"]; ok {
+	// 	config.Tty = value.(bool)
+	// }
+
+	// if value, ok := d["stdin_open"]; ok {
+	// 	config.OpenStdin = d.Get("stdin_open").(bool)
+	// }
 
 	if v, ok := d.GetOk("env"); ok {
 		config.Env = stringSetToStringSlice(v.(*schema.Set))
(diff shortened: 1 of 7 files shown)


docker/resource_docker_container_funcs.go
```diff
@@ -43,8 +43,18 @@ func resourceDockerContainerCreate(d *schema.ResourceData, meta interface{}) err
...
 	}
 
+	// if value, ok := d["tty"]; ok {
+	// 	config.Tty = value.(bool)
+	// }
+
+	// if value, ok := d["stdin_open"]; ok {
+	// 	config.OpenStdin = d.Get("stdin_open").(bool)
+	// }
+
```
suzuki-shunsuke, line 57:
```suggestion
```

suzuki-shunsuke:
Thank you for your contribution!

docker/structures_service.go
```diff
@@ -167,6 +167,12 @@ func flattenContainerSpec(in *swarm.ContainerSpec) []interface{} {
 	if len(in.Isolation) > 0 {
 		m["isolation"] = string(in.Isolation)
 	}
+	if in.TTY {
+		m["tty"] = in.TTY
+	}
+	if in.OpenStdin {
+		m["stdin_open"] = in.OpenStdin
+	}
```
suzuki-shunsuke, line 175: This function isn't used at `docker_container` resource, so I think we don't have to update this function in this pull request.

```suggestion
```

docker/resource_docker_container_v1.go
```diff
@@ -789,6 +789,16 @@ func resourceDockerContainerV1() *schema.Resource {
...
+			"tty": {
+				Type:     schema.TypeBool,
+				Default:  false,
+				Optional: true,
+			},
+			"stdin_open": {
+				Type:     schema.TypeBool,
+				Default:  false,
+				Optional: true,
+			},
```
suzuki-shunsuke, line 801: This function is used only for schema migration, so I think we don't have to update this function.

```suggestion
```

suzuki-shunsuke (changes requested):
Please update the document too.
https://github.com/kreuzwerker/terraform-provider-docker/blob/master/website/docs/r/container.html.markdown

docker/resource_docker_container_funcs.go
tanis80: Thank you. it was my first version, but I had some issue.
Tested again and works, updated as suggested

docker/structures_service.go
tanis80: Thank you. removed

docker/resource_docker_container_v1.go
tanis80: okay, removed as well

docker/resource_docker_container_funcs.go
```diff
@@ -43,17 +43,15 @@ func resourceDockerContainerCreate(d *schema.ResourceData, meta interface{}) err
...
+	if value, ok := d.GetOk("tty"); ok {
+		config.Tty = value.(bool)
+	}
 
-	// if value, ok := d["stdin_open"]; ok {
-	// 	config.OpenStdin = d.Get("stdin_open").(bool)
-	// }
+	if value, ok := d.GetOk("stdin_open"); ok {
+		config.OpenStdin = value.(bool)
+	}
```
suzuki-shunsuke: We can simply use `d.Get()` instead of `d.GetOk` in this case.
So this change is unneeded.

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema#ResourceData.Get
> If the key does exist in the schema but doesn't exist in the configuration, then the default value for that type will be returned. For strings, this is "", for numbers it is 0, etc.

scripts/testing/setup_private_registry.bat
```diff
@@ -9,7 +9,7 @@ call openssl req ^
   -x509 ^
   -days 365 ^
   -subj "/C=US/ST=Denial/L=Springfield/O=Dis/CN=127.0.0.1" ^
-  -keyout %~dp0certs\registry_auth.key ^
+  -keyout %~dp0certs\registry_authssss.key ^
```
suzuki-shunsuke, line 12: What's this change?

scripts/testing/setup_private_registry.bat
```diff
@@ -42,8 +42,8 @@ call docker run ^
   -e "REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd" ^
   -v %~dp0certs:/certs ^
   -e "REGISTRY_HTTP_TLS_CERTIFICATE=/certs/registry_auth.crt" ^
-  -e "REGISTRY_HTTP_TLS_KEY=/certs/registry_auth.key" ^
-  registry:2
+  -e "REGISTRY_HTTP_TLS_KEY=/certs/registry_authssss.key" ^
```
suzuki-shunsuke, line 45: What's this change?

tanis80: ops, sorry. my fault. I'll change back

tanis80: one note. I changed the image from registry:2 to registry:2.7.0. I was getting an error from Windows with registry:2.

suzuki-shunsuke: I don't use Windows, so unfortunately it is difficult for me to review this.
Would you tell us the error log?

tanis80: This was the issuse: docker local registry “exec: \”htpasswd\“: executable file not found in $PATH”

And I was find a solution from this articles in stackoverflow:

https://stackoverflow.com/questions/62531462/docker-local-registry-exec-htpasswd-executable-file-not-found-in-path

mavogel: Yes, we pinned the registry to `2.7.0` due to this error for unix as well. We'd probably think of doing test at `appveyor` @suzuki-shunsuke

suzuki-shunsuke (approved):

mavogel (approved):

PR #126, opened 2021-01-19
docs: fix legacy configuration style

suzuki-shunsuke:
> Warning: Interpolation-only expressions are deprecated
>
> Terraform 0.11 and earlier required all non-constant expressions to be
> provided via interpolation syntax, but this pattern is now deprecated. To
> silence this warning, remove the "${ sequence from the start and the }"
> sequence from the end of this expression, leaving just the inner expression.
>
> Template interpolation syntax is still used to construct strings from
> expressions when the template includes multiple interpolation sequences or a
> mixture of literal strings and interpolations. This deprecation applies only
> to templates that consist entirely of a single interpolation sequence.

We requires `Terraform >=0.12.x`, so we should fix legacy configuraiton style.

https://github.com/kreuzwerker/terraform-provider-docker#requirements

### DIFF
--- a/docker/resource_docker_container_test.go
+++ b/docker/resource_docker_container_test.go
@@ -960,12 +960,12 @@ func TestAccDockerContainer_multipleUploadContentsConfig(t *testing.T) {
 				
 				resource "docker_container" "foo" {
 					name     = "tf-test"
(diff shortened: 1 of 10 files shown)


mavogel (approved):
Yes, good catch

Issue #66, opened 2020-12-25
service secret/config file_mode interpreted as decimal instead of octal


mavogel:
<i>This issue was originally opened by @tomalok as https://github.com/hashicorp/terraform-provider-docker/issues/247. It was migrated here as a result of the community provider takeover from @kreuzwerker. The original body of the issue is below.</i><hr>
### Terraform Version
`Terraform v0.12.23`
`provider.docker: version = "~> 2.7"` (v2.7.0)

### Affected Resource(s)
Please list the resources as a list, for example:
- `docker_service`

### Terraform Configuration Files
```hcl
resource "docker_service "foo" {
  ...
  task_spec {
    ...
    container_spec {
      secrets {
        secret_id =
          secret_id   = "..."
          secret_name = "foo_pw__1"
          file_name   = "/run/secrets/foo_pw"
          file_uid    = "100"
          file_gid    = "101"
          file_mode   = "0440"
      }
    }
  }
}
```

### Plan Output
```
                secrets {
                    file_gid    = "101"
                    file_mode   = 440
                    file_name   = "/run/secrets/foo_pw"
                    file_uid    = "100"
                    secret_id   = "..."
                    secret_name = "foo_pw__1"
                }
```

### Expected Behavior
`file_mode` probably should have remained a string "0440" or have been converted from octal to decimal.

https://docs.docker.com/engine/reference/commandline/service_create/#create-a-service-with-secrets indicates that the secret's and config's `mode=` value should be a 4-number sequence, and explicitly shows a leading `0`.

### Actual Behavior
The integer `440` was used as the `file_mode` value, which corresponds to `0670` octal -- which is not the correct.

It's also interesting to note that `file_gid` and `file_uid` get preserved as strings, but `file_mode` does not.

### Temporary Workaround
Using a decimal value (i.e. `288` instead of `0440`) does the trick -- but this is counter-intuitive with the firmly-entrenched decades-old tradition of specifying mode in octal.

tomalok:
To my knowledge this hasn't yet been addressed.

suzuki-shunsuke:
I could reproduce the problem.

```console
$ terraform version
Terraform v0.14.9
+ provider registry.terraform.io/kreuzwerker/docker v2.11.0
```

```tf
terraform {
  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "2.11.0"
    }
  }
}

provider "docker" {
}

resource "docker_service" "foo" {
[...]

PR #163, opened 2021-03-30
fix: parse file_mode as 8 based value

suzuki-shunsuke:


internal/provider/helper.go
```diff
@@ -0,0 +1,21 @@
...
+	v := int(crc32.ChecksumIEEE([]byte(s)))
+	if v >= 0 {
+		return v
+	}
+	if -v >= 0 {
+		return -v
+	}
+	// v == MinInt
+	return 0
+}
```
suzuki-shunsuke, line 21: https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-hashcode-package

suzuki-shunsuke:
Hmm...
DiffSuppressFunc doesn't work as expected.
I'll investigate.

tomalok:
Per my comment in #66 this is still not working

internal/provider/file_mode.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	a, err := strconv.ParseInt(s, 8, 32)
+	if err != nil {
+		return 0, err
+	}
+	if a < 0 {
+		return 0, fmt.Errorf("file_mode must be greater equal than 0: %d", a)
+	}
+	if a > 0o777 {
+		return 0, fmt.Errorf("file_mode must be less equal than 0o777: %d", a)
```
tomalok, line 28: Limiting it to `0o777` means limits the mode to just `[ugo][rwx]` -- it should also be possible to set the setuid, setgid, and sticky bits.  To allow these, use `0o4777` as the maximum value.

Also, does `ParseInt(s, 8, 32)` handle the situation when the string is leads with `0` and `0o`?

Do we also want to keep the current behavior of assuming the string is a _decimal_ value, unless the string leads with a `0` or `0o`?  If not, we should make it very clear in the changelog (etc.) that the workaround will no longer work.

tomalok (changes requested):

mavogel:
@suzuki-shunsuke how about we skip the suppress function and make it a Breaking Change for `v3.0.0`? Would you be fine with this?

suzuki-shunsuke:
I'll fix the conflict.

> how about we skip the suppress function and make it a Breaking Change for v3.0.0? Would you be fine with this?

I'll consider.

mavogel:
Ok I'll remove it from the next milestone and we will put it accordingly to the changes to another `v2.x` or then to `v3.x`

PR #170, opened 2021-04-08
fix/move helpers

mavogel:
- [x] migrators
- [x] [flatteners](https://learn.hashicorp.com/tutorials/terraform/provider-create?in=terraform/providers#add-flattening-functions), expanders/creators/mappers


suzuki-shunsuke (approved):
👍

PR #178, opened 2021-04-19
feat: add pull_image setting to `docker_registry_image` datasource

### LINKED ISSUE #172: Local image not updated despite pull_triggers enabled
### Community Note

### Terraform (and docker Provider) Version

```
Terraform v0.14.10
+ provider registry.terraform.io/kreuzwerker/docker v2.11.0
```

### Affected Resource(s)

* `docker_image`

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    docker = {
      source = "kreuzwerker/docker"
      version = "2.11.0"
    }
  }
}

data "docker_registry_image" "test" {
  name = "saitho/test:latest"
}

resource "docker_image" "test" {
  name          = data.docker_registry_image.test.name
  pull_triggers = [data.docker_registry_image.test.sha256_digest]
  keep_locally = true
}

resource "docker_container" "test" {
  image = docker_image.test.latest
  name  = "test"
  restart = "unless-stopped"

  ports {
    external = 1337
    internal = 80
  }
}
```

### Debug Output

https://gist.github.com/saitho/93ec863cde73d2ad7ea04298f17c7592

### Panic Output

None.

### Expected Behaviour

I pushed a new version to the "latest" tag of my Docker image. A `terraform init` should pull the latest image and recreate the containers with that version.

### Actual Behaviour

While the container is recreated, the latest image is not pulled. Thus the old image is used for the new container.
In fact, running `docker pull` manually will pull the latest image. After that, `terraform apply` uses the latest version.

### Steps to Reproduce

1. Push a simple webserver image to Docker registry:
```docker
FROM nginx:alpine
RUN echo 'test' > /usr/share/nginx/html/index.html
```
2. Replace the image `saitho/test` in the Terraform file above with your image.
3. Deploy the Terraform file onto a server with Terraform.
4. `terraform init && terraform apply`. Approve application with "yes"
5. `curl http://localhost:1337` => returns "test"
6. Change the echo in the Dockerfile to print "test2" into the HTML file.
7. Build the image and push it to Docker registry
8. `terraform apply`. Approve application with "yes"
9. `curl http://localhost:1337` => should return "test2", but still returns "test"

### Important Factoids

None.

### References

None.

saitho:
Resolves #172 

Going by my suggestion I tried skipping the local image check.
I didn't know if it's possible to check if the _sha256_digest_ referenced by pull_triggers changed...

So I ended up adding a new setting "pull_image" to the `docker_registry_image` datasource which is _false_ per default. If enabled, it will pull the image if the _sha256_digest_ was modified during the read.

I can't assess whether this change affects other resources as well. Also I wasn't sure how to test that properly.
Please tell me what you think about this change. :+1:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,11 @@
 # Contributing
 
-By participating to this project, you agree to abide our [code of conduct](/CODE_OF_CONDUCT.md).
+By participating in this project, you agree to abide our [code of conduct](/CODE_OF_CONDUCT.md).
 
 ## Write Issue
 
 When you have a bug report or feature request or something, please create an issue from [here](https://github.com/kreuzwerker/terraform-provider-docker/issues/new/choose).
-Before creating an issue, please check whether same or releated issues exist.
+Before creating an issue, please check whether same or related issues exist.
 Please use issue templates as much as possible.
 
 ### Guide of Bug report
@@ -68,7 +68,7 @@ TF_LOG=INFO TF_ACC=1 go test -v ./internal/provider -run ^TestAccDockerImage_dat
 make testacc_cleanup
 ```
 
-Furthermore, we recommened running the linters for the code and the documentation:
+Furthermore, we recommend running the linters for the code and the documentation:
 
 ```sh
 # install all the dependencies
--- a/internal/provider/data_source_docker_registry_image.go
+++ b/internal/provider/data_source_docker_registry_image.go
@@ -31,12 +31,19 @@ func dataSourceDockerRegistryImage() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"pull_image": {
+				Type:     schema.TypeBool,
+				Default:  false,
+				Optional: true,
+			},
 		},
 	}
 }
 
 func dataSourceDockerRegistryImageRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	pullOpts := parseImageOptions(d.Get("name").(string))
+	imageName := d.Get("name").(string)
+	pullOpts := parseImageOptions(imageName)
 	authConfig := meta.(*ProviderConfig).AuthConfigs
 
 	// Use the official Docker Hub if a registry isn't specified
@@ -74,8 +81,21 @@ func dataSourceDockerRegistryImageRead(ctx context.Context, d *schema.ResourceDa
 		}
 	}
 
+	changed := d.Id() != digest
 	d.SetId(digest)
 	d.Set("sha256_digest", digest)
+
+	if d.Get("pull_image").(bool) && changed {
+		var data Data
+		// load local images into the data structure
+		client := meta.(*ProviderConfig).DockerClient
+		if err := fetchLocalImages(ctx, &data, client); err != nil {
+			return diag.Errorf("Got error when attempting to fetch local images: %s", err)
+		}
+		if err := pullImage(ctx, &data, client, authConfig, imageName); err != nil {
+			return diag.Errorf("Unable to pull image %s: %s", imageName, err)
+		}
+	}
 
 	return nil
 }
--- a/website/docs/d/registry_image.html.markdown
+++ b/website/docs/d/registry_image.html.markdown
@@ -16,7 +16,8 @@ to date on the latest available version of the tag.
 
 ```hcl
 data "docker_registry_image" "ubuntu" {
-  name = "ubuntu:precise"
+  name       = "ubuntu:precise"
+  pull_image = true
 }
 
 resource "docker_image" "ubuntu" {
@@ -30,6 +31,7 @@ resource "docker_image" "ubuntu" {
 The following arguments are supported:
 
 * `name` - (Required, string) The name of the Docker image, including any tags. e.g. `alpine:latest`
+* `pull_image` - (Optional, bool) Whether the image should be pulled when `sha256_digest` changed
 
 ## Attributes Reference
 


suzuki-shunsuke:
https://github.com/kreuzwerker/terraform-provider-docker/pull/178

```
Error: GPG private key required
```

https://github.com/kreuzwerker/terraform-provider-docker/blob/2845519dce58ff9f0564e93b2a2909e37873b28b/.github/workflows/compile.yaml#L25-L30

These secrets can't be available to builds of forks.

https://github.community/t/make-secrets-available-to-builds-of-forks/16166/33

suzuki-shunsuke:
Thank you for your contribution! 👍

suzuki-shunsuke:
This branch is out-of-date with the base branch. Please update the branch.

suzuki-shunsuke (changes requested):
Please update the document too.
https://github.com/kreuzwerker/terraform-provider-docker/blob/master/website/docs/d/registry_image.html.markdown

saitho:
> These secrets can't be available to builds of forks.

The job execution can be restricted to run only on the main repository:
```
if: github.repository == 'kreuzwerker/terraform-provider-docker'
```
Should I add that?

mavogel:
@saitho sorry for this, you're one of our first forks. Please rebase on the master. I fixed the issue in https://github.com/kreuzwerker/terraform-provider-docker/commit/ca2663562579a7094e1990c7cbeb48cb59ab1c85 because for the `compile` action we don't need the signing

saitho:
> @saitho sorry for this, you're one of our first forks. Please rebase on the master. I fixed the issue in [ca26635](https://github.com/kreuzwerker/terraform-provider-docker/commit/ca2663562579a7094e1990c7cbeb48cb59ab1c85) because for the `compile` action we don't need the signing

Alright, no worries :) I rebased, pipeline looks good now.

mavogel:
@suzuki-shunsuke please re-review :)

suzuki-shunsuke (changes requested):
> I didn't know if it's possible to check if the sha256_digest referenced by pull_triggers changed...

I think we can use [schema#ResourceData.HasChange](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema#ResourceData.HasChange).
So I think we don't have to add a new attribute `pull_image`.

suzuki-shunsuke:
> I think we can use schema#ResourceData.HasChange.

Oh, I found a problem.
`ForceNew` of `pull_triggers` is `true`, so we can't check if `pull_triggers` is changed.
When `pull_triggers` is changed, the `docker_image` resource would be recreated.
If `docker_image`'s `keep_locally` is `false`, the old image is removed from local and the image digest is updated,
but if `keep_locally` is `true` the old image isn't removed so the image digest isn't updated.

The workaround is to make `keep_locally` false.

To solve this problem, I guess we have to change `ForceNew` of `pull_triggers` to `false`.

suzuki-shunsuke:
I changed `ForceNew` of `pull_triggers` to `false` in #184 .
Then the local image digest was updated properly, but the other problem was found.
When `terraform plan` is run, even if a docker_image resource is updated, the dependent docker_container resource isn't updated.
I guess the cause is [Computed](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed) of `docker_image.latest` is `true`.

mavogel:
@saitho and @suzuki-shunsuke are you still on this?

saitho:
I'm still interested in getting this solved. However I'm not too much into Terraform to solve the need-to-apply-twice issue in the PR.

suzuki-shunsuke:
> @saitho and @suzuki-shunsuke are you still on this?

No. I have no idea to solve the problem.

* https://github.com/kreuzwerker/terraform-provider-docker/pull/178#issuecomment-826062175
* https://github.com/kreuzwerker/terraform-provider-docker/pull/178#issuecomment-826069637

Do you have any idea?

stephenliberty:
$.02:

We got bit by the related bug (#172). After mulling this over for a bit, I think you might be trying to solve a use case that was created by accident, rather than the one that users really want. Maybe. This is a long and slightly opinionated rant, I'm sorry.

Right now it seems that the bug is as follows (for sake of clarity of conversation):
1. Create docker_registry_image resource, let's say `nginx:latest`
2. Create docker_image resource, pull trigger set to registry_image, keep_locally set to true
3. Upstream registry image changes, but local image does not change

I don't think the registry image is the issue here - I think it's performing just fine and shouldn't be touched. I also think that the local image is working fine too! We're just telling the provider to drop "latest" onto the machine, and not to delete the image if the resource goes away, right? When the upstream sha changes and it's recreated, the provider finds that the latest tag already exists, and instead of forcibly doing anything with that image, it simply says "cool, image tag is here, requested state matches". If we wanted something more specific, wouldn't it be better to.. well.. be specific?

Again, for sake of conversation, the situation is easily reproducible via the following:

```bash
docker pull nginx:1.15.3-alpine
docker tag nginx:1.15.3-alpine nginx:latest
```
```hcl
terraform {
  required_providers {
    docker = {
      source = "kreuzwerker/docker"
      version = "2.15.0"
    }
  }
}

data "docker_registry_image" "nginx" {
  name = "nginx:latest"
}

resource "docker_image" "nginx" {
  keep_locally = true
  name          = data.docker_registry_image.nginx.name
  pull_triggers = [data.docker_registry_image.nginx.sha256_digest]
}
```
```bash
terraform apply
<snip>
docker_image.nginx: Creation complete after 0s [id=sha256:994032453556b56420d66d53b7d8db1a74e1193165e2a070e50f533d849d9833nginx:latest]
```
Even though nginx 1.15 is not even close to latest, the state that we've requested does indeed exist - there is a local docker_image with the tag nginx:latest, so everything immediately succeeds - the output resource is the sha256 of 1.15 (verify: https://hub.docker.com/layers/nginx/library/nginx/1.15.3-alpine/images/sha256-994032453556b56420d66d53b7d8db1a74e1193165e2a070e50f533d849d9833?context=explore). It doesn't matter if the _actual_ `latest` sha updates, because the state we're expecting still exists.

If we really wanted the always-latest-image, using the following would likely net a better result:
```hcl
//same as above, with this change:
resource "docker_image" "nginx" {
  keep_locally = true
  name          = "nginx@${data.docker_registry_image.nginx.sha256_digest}"
  pull_triggers = [data.docker_registry_image.nginx.sha256_digest]
}
```
```bash
terraform apply
<snip>
docker_image.nginx: Modifications complete after 10s [id=sha256:e9ce56a96f8e0e9f75051f258a595d1257bd6bb91913b79455ea77e67e686c5cnginx@sha256:d1ce0f99f6a8acc9162c29497014716c44d126f1d41deee40a2c13e3d9d9b02a]
```
As of 1am pacific time this morning, https://hub.docker.com/layers/nginx/library/nginx/latest/images/sha256-eb3af4f7de38aae4ed25e1ca986c2d684d1d6b0ab470b5ced7ef41d864b66aa0?context=explore was the `nginx:latest` sha. So, in this case, I'm getting _exactly_ what I expected - as of the time of run, I got the latest version of nginx downloaded. Which is likely the actual use case.

My **opinion** is that instead of trying to fix this behavior, you give two new, backwards compatible items and update the docs:

1. add another output that gives an easily digestible link to the image. In cases where an image is passed in by variable, trying to parse that out would be a pain - being able to say, for instance, `data.docker_registry_image.nginx.pull_uri` would be incredibly helpful.
2. add a `tag` resource that would allow tag creation, so if there is a use case for actually making sure that there is a (for example) `nginx:latest` on a machine and that it's kept up to date even with `keep_locally` set to true, you could do that. Maybe something like..
```hcl
docker_tag "nginx_latest" {
   tag = "nginx:latest"
   from = docker_image.nginx.repo_digest
}
```

PR #182, opened 2021-04-22
fix: replace for loops with StateChangeConf

mavogel:
- remove all for loop to wait for a state and add StateChangeConf for this
- added local acceptance setup with Dockerfile for easier debugging of ubuntu based errors when developing on another OS like OSX


mavogel:
@suzuki-shunsuke you can ignore the waiting checks, as I updated the matrix for `acc_test` and the _waiting_ ones are from the old config. Feel free to review

PR #190, opened 2021-05-09
Win test spaces

Lexy2:
Improved compatibility with Docker for Windows (Linux containers).

- Added support for paths with spaces
- Updated the testing script configuration to match Linux shell test scripts
- Added support for backslash path separator in acceptance tests + escaping such paths in terraform files
- Fixed a couple of bugs in provider functions that had prevented tests from passing

### DIFF
--- a/internal/provider/resource_docker_container_funcs.go
+++ b/internal/provider/resource_docker_container_funcs.go
@@ -467,9 +467,10 @@ func resourceDockerContainerCreate(ctx context.Context, d *schema.ResourceData, 
 
 	if d.Get("attach").(bool) {
 		var b bytes.Buffer
-
+		logsRead := make(chan bool)
 		if d.Get("logs").(bool) {
 			go func() {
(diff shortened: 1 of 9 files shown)


mavogel:
Thanks a lot, @Lexy2 for the PR :) WDYT about adding matrix steps for `windows-latest` platform version as well and adapting the steps in CI?

Lexy2:
Thanks, Manuel, I'll take a look and update the PR.

mavogel:
Ok thanks for the detailed explanation. Yes then we keep it like this for the moment

Lexy2:
Thanks. How do I do that, formally?

mavogel:
Good to know. When this is merge you're a collaborator and for the future this will be solved then :)

suzuki-shunsuke:
I think collaborator is different from contributor.

mavogel (approved):
LGTM. Thanks a lot :)

PR #193, opened 2021-05-13
feat/doc generation

mavogel:
## Details
- enables the generation of the documentation from the descriptions within the schema based on https://github.com/hashicorp/terraform-plugin-docs
- there are still bugs/missing features 
   - for rendering nested resources in data-source. See https://github.com/hashicorp/terraform-plugin-docs/issues/64
   - for default values: https://github.com/hashicorp/terraform-plugin-docs/issues/65
   - for rendering `read-only` blocks: https://github.com/hashicorp/terraform-plugin-docs/issues/66

## TODOs
- [x] add all descriptions for the attributes
- [x] how many attributes are required for datasources?
- [x] customize all templates to make descriptions more concise 
- [x] read through all docs if they make sense and are understandable

## Questions: 
- [x] do we still need the link check for markdown? -> fixed rules in fb33652


suzuki-shunsuke (commented):
How about running `make website-generation` and checking if there is any change in CI?

suzuki-shunsuke:
Thank you for your great work ❤️

mavogel:
@suzuki-shunsuke thanks for the hint. Added it! Take a look at the latest build/action

suzuki-shunsuke (approved):

PR #161, opened 2021-03-29
feat: support to import an image with image name

suzuki-shunsuke:

Issue #156, opened 2021-03-26
Change `docker_image` resource id to image id and support import


suzuki-shunsuke:
* resource: `docker_image`
* attribute: `id`

## AS IS

https://github.com/kreuzwerker/terraform-provider-docker/blob/6b0f977735dbc2ff99faa9278f4d51f2bf077b02/internal/provider/resource_docker_image_funcs.go#L75

The value of id is `${image digest}${image name}`.

e.g. `sha256:bfba26ca350c153700592ebd7e4613f25a7a8779289e041eb68a97623277de82nginx:1.17.0-alpine`

```json
          "attributes": {
            "build": [],
            "force_remove": null,
            "id": "sha256:bfba26ca350c153700592ebd7e4613f25a7a8779289e041eb68a97623277de82nginx:1.17.0-alpine",
            "keep_locally": null,
            "latest": "sha256:bfba26ca350c153700592ebd7e4613f25a7a8779289e041eb68a97623277de82",
            "name": "nginx:1.17.0-alpine",
            "output": null,
            "pull_trigger": null,
            "pull_triggers": null
          }
```

The format is dirty and it is difficult to import the resource.

## TO BE

The id should be the image id, and `docker_image` can be imported by the name attribute.

PR #212, opened 2021-06-01
fix/service image name

mavogel:
- remove image name suppress function which only caused confusion
- update examples and test to use the `docker_image` resource to determine the `sha256@<hash>` digest for a image
- adds a `docker_image` datasource


docs/resources/service.md
```diff
@@ -73,12 +78,17 @@ resource "docker_network" "test_network" {
   driver = "overlay"
 }
 
+resource "docker_image" "foo" {
```
tomalok, line 81: if the image already exists somewhere, would this be better represented as a `data` source?  Or is the idea that having it be a `resource` will check whether the image:tag has a new hash associated with it?

docs/resources/service.md
```diff
@@ -24,12 +24,17 @@ The following configuration starts a Docker Service with
...
+  keep_locally = true
+}
+
 resource "docker_service" "foo" {
   name = "foo-service"
 
   task_spec {
     container_spec {
-      image = "repo.mycompany.com:8080/foo-service:v1"
+      image = docker_image.foo.latest
```
tomalok, line 37: is there a way to avoid confusion between the image's `:latest` tag and the `.latest` attribute, which in this case does not appear to be the same.

mavogel: yes, you're right. This is confusing. I added a more extended description to it but we will also revise it in #161

docs/resources/service.md
mavogel: This could also be an option. ATM there is not data-source for an image. Yes, the `pull_triggers` will check for a newer version of the image

mavogel: I added a `docker_image` datasource in a3a236145b62bee0f28ea1a904ae988cd3285dcb

docs/resources/service.md
mavogel: The description got updated. we might remove the `latest` attribute and replace it with a clearer one like `sha256_digest`

tomalok: yep, that'd be much better.  :)

internal/provider/data_source_docker_image_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var imageDigestRegexp = regexp.MustCompile(`^sha256:[A-Fa-f0-9]+$`)
+var imageNameWithTagAndDigestRegexp = regexp.MustCompile(`^.*:.*@sha256:[A-Fa-f0-9]+$`)
+
+func TestAccDockerImageDataSource_withSpecificTag(t *testing.T) {
```
tomalok, line 20: Is it possible to also test references to non-existent image/tag/sha256?

docs/resources/image.md
```diff
@@ -80,7 +80,7 @@ resource "docker_image" "zoo" {
 
 ### Read-Only
 
-- **latest** (String) The ID of the image.
+- **latest** (String) The ID of the image in the form of `sha256:<hash>` image digest. Do not confuse it with the default `latest` tag.
```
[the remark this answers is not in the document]
mavogel: Good point. Will add this and remove it in `v3.0.0`

internal/provider/data_source_docker_image_test.go
mavogel: yes, I can add such a test

docs/resources/service.md
mavogel: added in 812ae724f2045cf38b68917f82dec443c764cfd2

docs/resources/image.md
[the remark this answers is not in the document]
mavogel: done in 812ae724f2045cf38b68917f82dec443c764cfd2

internal/provider/data_source_docker_image_test.go
mavogel: done in 86f24ca6153a783fe77ae3bcd160ce70efe463c9

mavogel:
@tomalok thanks for all you input :) Let me know if there is anything else from your side. Feel free to close the conversations if they are resolved for you

internal/provider/data_source_docker_image.go
```diff
@@ -0,0 +1,62 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The name of the Docker image, including any tags or SHA256 repo digests.",
+				Required:    true,
+			},
+			"latest": {
+				Type:        schema.TypeString,
+				Description: "The ID of the image in the form of `sha256:<hash>` image digest. Do not confuse it with the default `latest` tag.",
```
suzuki-shunsuke, line 25: In my understanding, the image digest is created when we push the image to the Docker Registry,
which means if the image isn't pushed to the Docker Registry, the image doesn't have image digest.
So I don't know whether it is good to deprecate the image id.

internal/provider/data_source_docker_image.go
```diff
@@ -0,0 +1,62 @@
...
+
+	foundImage := searchLocalImages(ctx, client, data, imageName)
+	if foundImage == nil {
+		return diag.Errorf("did not find docker image '%s'", imageName)
+	}
+
+	d.SetId(foundImage.ID)
+	d.Set("name", imageName)
+	d.Set("latest", foundImage.ID)
+	d.Set("sha256_digest", foundImage.ID)
```
suzuki-shunsuke, line 59: This is not an image digest but an image id.

suzuki-shunsuke: I guess ImageSummary doesn't have an image digest although I don't check.
https://pkg.go.dev/github.com/docker/docker@v20.10.6+incompatible/api/types#ImageSummary

internal/provider/resource_docker_image.go
```diff
@@ -22,7 +22,14 @@ func resourceDockerImage() *schema.Resource {
 
 			"latest": {
 				Type:        schema.TypeString,
-				Description: "The ID of the image.",
+				Description: "The ID of the image in the form of `sha256:<hash>` image digest. Do not confuse it with the default `latest` tag.",
```
suzuki-shunsuke, line 25: It looks a breaking change.

internal/provider/data_source_docker_image.go
tomalok: The deprecation should be just renaming the confusing `latest` attribute.

If I've done a `docker build` on the local machine, and the image is created, it should still have a value for this.

internal/provider/data_source_docker_image.go
tomalok: This is a good point -- while it may appear to be a sha256 digest (and maybe that's how it's generated from some collection of uniquely identifying data), it's still simply an `id`.

If `d.SetId(foundImage.ID)` also sets the `id` attribute, why not use that instead of `latest` or `sha256_digest` altogether, if it does indeed uniquely identify the image?

tomalok:
> @tomalok thanks for all you input :) Let me know if there is anything else from your side. Feel free to close the conversations if they are resolved for you

I appear to be unable to close conversations?  (or maybe I'm just not finding the button to do so...)

README.md
```diff
@@ -42,9 +42,38 @@ resource "docker_image" "nginx" {
...
+  }
+}
+
+# Or create a service resource
+# -> same as 'docker service create -d -p 8081:80 --name nginx-service --replicas 2 nginx:latest'
+resource "docker_service" "nginx_service" {
+  name = "nginx-service"
+  task_spec {
+    container_spec {
+      image = docker_image.nginx.latest
```
tomalok, line 62: Deprecated `latest` attribute.  Based on what I'm seeing in https://github.com/kreuzwerker/terraform-provider-docker/blob/fix/service-image-name/internal/provider/data_source_docker_image.go, a simple `docker_image.nginx.id` should work the same way, and we don't need the new `sha256_digest` there, either.

How does using the local image `id` or registry image `sha256_digest` work versus using a `<registry>/<image>:<tag>[@<sha256>]` hash, when it comes to a docker service, if the task is scheduled on a node that the provider isn't currently communicating with (where, presumably the `docker_image` or `docker_registry_image` would be located)?

I think it's the local node's docker engine that does the pull from the registry if it's not locally available.  Having it try to pull a hash probably doesn't give it all the information it needs...
```
m1:~$ docker image inspect 075d847609aa
[
    {
        "Id": "sha256:075d847609aa84dbbe8a5a36ca3852dd75bbcd643a92bb5e1757e75727d44210",
        "RepoTags": [
            "jakesys/aws2:latest"
        ],
        "RepoDigests": [
            "jakesys/aws2@sha256:aa579322e9a0a27c5bece3cc9f5a76e5482c2c846e555917f17ce28034a78d3a"
        ],
... 
]
```
Perhaps what we really want to provide here is the image's `RepoDigest` string.

docs/data-sources/image.md
```diff
@@ -0,0 +1,53 @@
...
+- **name** (String) The name of the Docker image, including any tags or SHA256 repo digests.
+
+### Optional
+
+- **id** (String) The ID of this resource.
+
+### Read-Only
+
+- **latest** (String, Deprecated) The ID of the image in the form of `sha256:<hash>` image digest. Do not confuse it with the default `latest` tag.
+- **sha256_digest** (String) The image sha256 digest in the form of `sha256:<hash>`.
```
tomalok, line 51: Having a `repo_digest` string would probably be the most useful, rather than repeating the `id` value again, which as @suzuki-shunsuke points out is not an actual image digest (but the value does appear to be stable between two different machines).

docs/resources/image.md
```diff
@@ -80,8 +80,9 @@ resource "docker_image" "zoo" {
 
 ### Read-Only
 
-- **latest** (String) The ID of the image.
+- **latest** (String, Deprecated) The ID of the image in the form of `sha256:<hash>` image digest. Do not confuse it with the default `latest` tag.
 - **output** (String, Deprecated)
+- **sha256_digest** (String) The image sha256 digest in the form of `sha256:<hash>`.
```
tomalok, line 85: Full `repo_digest` string instead?

docs/resources/service.md
tomalok: As mentioned in the top-level README.md comment, we probably need to provide the `docker_service.task_spec.image_spec.image` attribute either a full `repo_digest` string or the usual `<registry>/<image>[:<tag>]` so that other swarm nodes that the provider's not connected to know what they need to pull.

docs/resources/service.md
```diff
@@ -73,12 +105,17 @@ resource "docker_network" "test_network" {
...
+  name         = "repo.mycompany.com:8080/foo-service:v1"
+  keep_locally = true
+}
+
 resource "docker_service" "foo" {
   name = "tftest-service-basic"
 
   task_spec {
     container_spec {
-      image = "repo.mycompany.com:8080/foo-service:v1"
```
tomalok, line 81: you may want to keep an example of specifying an image this way, too?

internal/provider/data_source_docker_image.go
tomalok: I've confirmed that although image id is stable from machine to machine, it's not actually the same as the image digest itself -- probably includes some extra metadata when the id is generated.

internal/provider/data_source_docker_image.go
tomalok: If we have (or can construct) the `repo_digest`, that'd be the best thing to have here -- especially since `docker_service` will need to pass all that information along to the nodes it gets scheduled on, so they know what to pull.
```
m1:~$ docker image inspect 075d847609aa
[
    {
        "Id": "sha256:075d847609aa84dbbe8a5a36ca3852dd75bbcd643a92bb5e1757e75727d44210",
        "RepoTags": [
            "jakesys/aws2:latest"
        ],
        "RepoDigests": [
            "jakesys/aws2@sha256:aa579322e9a0a27c5bece3cc9f5a76e5482c2c846e555917f17ce28034a78d3a"
        ],
...
}

File: internal/provider/resource_docker_service.go [hunk 12]
```diff
@@ -83,10 +79,9 @@ func resourceDockerService() *schema.Resource {
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"image": {
-										Type:             schema.TypeString,
-										Description:      "The image name to use for the containers of the service",
-										Required:         true,
-										DiffSuppressFunc: suppressIfSHAwasAdded(),
+										Type:        schema.TypeString,
+										Description: "The image name to use for the containers of the service. Use the `docker_image` resource for this, as shown in the examples. Although direct image names like `nginx:latest` works it is not recommend to trigger updates",
```
tomalok, line 83: Should be able to use `docker_image` or `docker_registry_image` (from resource or data source) as long as we can get a full "repo_digest" string from each of them.

(Local `docker_image`s that haven't been pushed to a repository, will require that same image to exist on all swarm nodes that will run the service.)

internal/provider/data_source_docker_image.go
mavogel: Good point. What confused me was that `RepoDigests` is an array. So what happens if there a multiple? Shall we take the first?

internal/provider/resource_docker_image.go
mavogel: It's not. I still use the `ID` like before. Just update the description, which is misleading. Although the `ID` has the same form as the digest, it's simply an `ID`

README.md
mavogel: > a simple docker_image.nginx.id should work the same way, and we don't need the new sha256_digest there, either.

Yes, that's true.

> Having it try to pull a hash probably doesn't give it all the information it needs...

Good point. I did not think of all the swarm node. So either 
- `<registry>/<image>:<tag>` or
- `<registry>/<image>@<sha256>` or
- `<registry>/<image>:<tag>@<sha256>`
are the ways to go

internal/provider/resource_docker_image.go
suzuki-shunsuke: I see. Thank you for your explanation.

internal/provider/data_source_docker_image.go
tomalok: Taking a closer look at what multiple `RepoDigests` means, I've managed to get multiple values...
```
jake@jimini ~ % docker image ls                                                     
REPOSITORY                                TAG               IMAGE ID       CREATED        SIZE
jakesys/aws2                              latest            a406618724ac   10 hours ago   95.7MB
registry.gitlab.com/jakesys/aws2/master   2.2.9             a406618724ac   10 hours ago   95.7MB
registry.gitlab.com/jakesys/aws2/master   latest            a406618724ac   10 hours ago   95.7MB
jakesys/aws2                              <none>            7f92c8cc54e5   2 days ago     95.7MB
```
...because the same image can be in _multiple_ registries.  The untagged image was formerly the `latest`, the new `latest` was built by GitLab CI last night, and pushed to hub.docker.com.  I also pulled the corresponding version tag from the GitLab registry.
```
jake@jimini ~ % docker image inspect jakesys/aws2
[
    {
        "Id": "sha256:a406618724ac720e08c077b604c25946f6c5da1702f5e0b30f72c91d1e40643f",
        "RepoTags": [
            "jakesys/aws2:latest",
            "registry.gitlab.com/jakesys/aws2/master:2.2.9",
            "registry.gitlab.com/jakesys/aws2/master:latest"
        ],
        "RepoDigests": [
            "jakesys/aws2@sha256:3734f6bf3c8bd2a384caba078a4e4913f8809f871b473978bbf27cf726977b1c",
            "registry.gitlab.com/jakesys/aws2/master@sha256:3734f6bf3c8bd2a384caba078a4e4913f8809f871b473978bbf27cf726977b1c"
        ],
...
  }
]
```

The `RepoDigest` that matches the image name being used to define the `docker_registry_image` data source (or the Terraform-managed resource) would be the one to use.

README.md
tomalok: While tracking down what's in `RepoDigests`, I discovered that this is _empty_ for locally-built images that have not been pushed to a registry -- IIRC, @suzuki-shunsuke already alluded to this.

So, for references to local `docker_image` data sources and references, the use of `id` should be used, too.

internal/provider/data_source_docker_image.go
mavogel: > So I don't know whether it is good to deprecate the image id.

I'd suggest to remove the `latest` attribute and encourage folks to use `id` or `sha256_digest`

README.md
mavogel: See 18c3f355c27c5a82480f3eb9a5053144adcc7f4f

docs/resources/image.md
mavogel: Yes :) See 18c3f355c27c5a82480f3eb9a5053144adcc7f4f

docs/data-sources/image.md
```diff
@@ -0,0 +1,53 @@
...
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- **name** (String) The name of the Docker image, including any tags or SHA256 repo digests.
+
+### Optional
+
+- **id** (String) The ID of this resource.
```
suzuki-shunsuke, line 46: It looks strange. `id` is Read Only.
This document is generated by `terraform-plugin-docs`, so this is a problem of `terraform-plugin-docs` 🤔

testdata/resources/docker_service/updateFailsAndRollbackConvergeConfig.tf
```diff
@@ -5,12 +5,25 @@ provider "docker" {
...
+  pull_triggers = [data.docker_registry_image.tftest_image.sha256_digest]
+}
+
 resource "docker_service" "foo" {
   provider = "docker.private"
   name     = "tftest-service-updateFailsAndRollbackConverge"
   task_spec {
     container_spec {
-      image             = "%s"
+      image             = docker_image.tftest_image.latest
```
suzuki-shunsuke, line 26:
```suggestion
      image             = docker_image.tftest_image.id
```

testdata/resources/docker_service/testAccDockerServicePrivateImageConverge.tf
```diff
@@ -1,14 +1,29 @@
...
+}
+
 resource "docker_service" "foo" {
-  name = "tftest-service-foo"
+  provider = "docker.private"
+  name     = "tftest-service-foo"
   task_spec {
     container_spec {
-      image             = "%s"
+      image             = docker_image.tftest_image.latest
```
suzuki-shunsuke, line 26:
```suggestion
      image             = docker_image.tftest_image.id
```

testdata/resources/docker_service/testAccDockerServicePartialReplicationConfigStep3.tf
```diff
@@ -4,11 +4,17 @@ provider "docker" {
...
+  keep_locally = false
+  force_remove = true
+}
+
 resource "docker_service" "foo" {
   name = "tftest-service-basic"
   task_spec {
     container_spec {
-      image             = "127.0.0.1:15000/tftest-service:v1"
+      image             = docker_image.tftest_image.latest
```
suzuki-shunsuke, line 17:
```suggestion
      image             = docker_image.tftest_image.id
```

internal/provider/data_source_docker_image.go
suzuki-shunsuke: Looks good.
The the fallback should be documented.

tomalok: I'm going to disagree here.  If the local `docker_image` data source / resource does not have these values set, a `null` value would be more appropriate to indicate that the image was not pulled from or pushed to a registry.  Also the image's `ID` string has a completely different format than the `repoDigest`.

We shouldn't need to fill in all the empty blanks...  ;)

Certainly we should document what could cause a `null` (or empty string) value for the read-only attribute.

Fleshing out the `repo_digest` attribute on the `docker_registry_image` data source / resource shouldn't have this issue... :)

docs/data-sources/image.md
mavogel: Yes it is :) see https://github.com/hashicorp/terraform-plugin-docs/issues/45

testdata/resources/docker_service/testAccDockerServicePartialReplicationConfigStep3.tf
mavogel: see 58a5e8599beaabff8aedab2682e1e894dbaddc93 and 3a6790cf1eb69279e2a0b37ad49d8a2d89f2c5f2

testdata/resources/docker_service/testAccDockerServicePrivateImageConverge.tf
mavogel: see 58a5e8599beaabff8aedab2682e1e894dbaddc93 and 3a6790cf1eb69279e2a0b37ad49d8a2d89f2c5f2

testdata/resources/docker_service/updateFailsAndRollbackConvergeConfig.tf
mavogel: see 58a5e8599beaabff8aedab2682e1e894dbaddc93 and 3a6790cf1eb69279e2a0b37ad49d8a2d89f2c5f2

examples/resources/docker_service/resource-basic-image-ds.tf
```diff
@@ -0,0 +1,19 @@
+data "docker_image" "foo" {
+  name = "repo.mycompany.com:8080/foo-service:v1"
+}
+
+resource "docker_service" "foo" {
+  name = "foo-service"
+
+  task_spec {
+    container_spec {
+      image = data.docker_image.foo.latest
```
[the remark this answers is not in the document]
mavogel: see 58a5e8599beaabff8aedab2682e1e894dbaddc93 and 3a6790cf1eb69279e2a0b37ad49d8a2d89f2c5f2

docs/resources/service.md
```diff
@@ -48,6 +48,33 @@ The following command is the equivalent:
...
+data "docker_image" "foo" {
+  name = "repo.mycompany.com:8080/foo-service:v1"
+}
+
+resource "docker_service" "foo" {
+  name = "foo-service"
+
+  task_spec {
+    container_spec {
+      image = data.docker_image.foo.latest
```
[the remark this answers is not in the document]
mavogel: see 58a5e8599beaabff8aedab2682e1e894dbaddc93 and 3a6790cf1eb69279e2a0b37ad49d8a2d89f2c5f2

internal/provider/data_source_docker_image.go
mavogel: yes, see 9ed8378f3106cbb2c8f7a5a7c974f3419a74f9fb

mavogel: Yes, I removed it in 9ed8378f3106cbb2c8f7a5a7c974f3419a74f9fb

README.md
mavogel: > So, for references to local docker_image data sources and references, the use ofid should be allowed, too.

and `repo_digest` as well

internal/provider/data_source_docker_image.go
mavogel: See d04ff80e1808dd18adfaf3a9354947a1853d3759 and the docs in 5ec1577bdbba9cbce5ae13812a219741013b4628

tomalok: this is better -- but there's still one more edge case where the image was pulled from somewhere, pushed somewhere else, but the tag being referenced in the `data` is that local-only tag...

i think the following would cover all cases...
* empty `RepoDigests`, return `""`
* one `RepoDigests`, return it
* multiple `RepoDigests`...
  * one matches the `imageName`, return it
  * nothing matches the `imageName`, return the first one *(new edge case)*

That should cover all situations.

mavogel: Alright, I added a fix in 2c5d6f8e48a805cfa38420de3fce594466fecc6f and did throw an error if _nothing matches_

internal/provider/resource_docker_service.go
mavogel: I adapted the description in 4d0f69d7acffb77f88e450d39bb002e2b52633a5 and tested the cases locally

internal/provider/data_source_docker_image.go
tomalok: Looks good, thanks!

testdata/resources/docker_service/updateFailsAndRollbackConvergeConfig.tf
```diff
@@ -5,12 +5,25 @@ provider "docker" {
...
+  provider             = "docker.private"
+  name                 = "%s"
+  insecure_skip_verify = true
+}
+resource "docker_image" "tftest_image" {
+  provider      = "docker.private"
+  name          = data.docker_registry_image.tftest_image.name
+  keep_locally  = false
+  force_remove  = true
+  pull_triggers = [data.docker_registry_image.tftest_image.sha256_digest]
```
tomalok, line 18: Does the `docker_registry_image`'s `sha256_digest` value contain `sha256:` like `RepoDigests`? -- if so, it's conceivable that at some point a different hash might be used?

internal/provider/resource_docker_image_funcs.go
```diff
@@ -63,9 +63,12 @@ func resourceDockerImageRead(ctx context.Context, d *schema.ResourceData, meta i
 		return nil
 	}
 
+	repoDigest := determineRepoDigest(imageName, foundImage)
+
 	// TODO mavogel: remove the appended name from the ID
 	d.SetId(foundImage.ID + d.Get("name").(string))
```
tomalok, line 69: time to remove the appended name?

internal/provider/resource_docker_image.go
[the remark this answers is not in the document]
mavogel: is does have a `Deprecated` attribute

internal/provider/resource_docker_image_funcs.go
mavogel: we will tackle this in #156

README.md
```diff
@@ -42,9 +42,38 @@ resource "docker_image" "nginx" {
 }
 
 # Create a docker container resource
-# -> same as 'docker run --name nginx -d nginx:latest'
+# -> same as 'docker run --name nginx -p8080:80 -d nginx:latest'
 resource "docker_container" "nginx" {
   name    = "nginx"
   image   = docker_image.nginx.latest
```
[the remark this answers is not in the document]
mavogel: we will tackle this in #156 . There is more to do under the hood.

testdata/resources/docker_service/updateFailsAndRollbackConvergeConfig.tf
mavogel: Yes it also contains the `sha256:` but without a leading container name, like `nginx` in this case
```
data_docker_image_nginx_repo_digest            = "nginx@sha256:36b74457bccb56fbf8b05f79c85569501b721d4db813b684391d63e02287c0b2"
data_docker_registry_image_nginx_sha256_digest = "sha256:36b74457bccb56fbf8b05f79c85569501b721d4db813b684391d63e02287c0b2"
```

README.md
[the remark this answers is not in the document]
mavogel: done in 6593c62545d671b0cbb9b33ec0ca9f4d0224fb04

testdata/resources/docker_service/testAccDockerServiceConvergeAndStopGracefully.tf
```diff
@@ -4,11 +4,16 @@ provider "docker" {
...
+  name         = "127.0.0.1:15000/tftest-service:v1"
+  keep_locally = false
+}
+
 resource "docker_service" "foo" {
   name = "tftest-service-basic-converge"
   task_spec {
     container_spec {
-      image             = "127.0.0.1:15000/tftest-service:v1"
+      image             = docker_image.tftest_image.latest
```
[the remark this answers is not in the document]
mavogel: done in 132b9f46b1d9463052a1c826f635c780438ef175

testdata/resources/docker_service/testAccDockerServiceFullSpec.tf
```diff
@@ -33,7 +38,7 @@ resource "docker_service" "foo" {
 
   task_spec {
     container_spec {
-      image = "127.0.0.1:15000/tftest-service:v1"
+      image = docker_image.tftest_image.latest
```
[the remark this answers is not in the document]
mavogel: done in 132b9f46b1d9463052a1c826f635c780438ef175

testdata/resources/docker_service/testAccDockerServicePrivateImageConverge.tf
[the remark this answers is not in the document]
mavogel: done in 132b9f46b1d9463052a1c826f635c780438ef175

testdata/resources/docker_service/updateFailsAndRollbackConvergeConfig.tf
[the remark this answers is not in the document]
mavogel: done in 132b9f46b1d9463052a1c826f635c780438ef175

internal/provider/resource_docker_image.go
```diff
@@ -22,7 +22,14 @@ func resourceDockerImage() *schema.Resource {
...
+				Computed:    true,
+				Deprecated:  "Use sha256_digest instead",
```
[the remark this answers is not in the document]
mavogel: done in 0baffdffaec9c91ca96ddbdf01e5023975663c37

suzuki-shunsuke (approved):

PR #223, opened 2021-06-09
internal/provider: include image name in error message

kevinburkemeter:
Without this, if the Docker image is not found, it's difficult to know
which image exactly is 404'ing - you get a very generic error message,
which can make it difficult to spot typos or other problems with the
image name.


mavogel (approved):
Thank you!

PR #241, opened 2021-06-30
feat(service): add alias for networks

mavogel:
Add the ability to pass an `alias` for the containers in the network. Currently, the `network_advanced` workaround does not work properly.


Junkern:
Added `driver_opts` to the `network_advanced` block.
According to https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-attribute-removal will release this as a new minor version and then for the next major version will remove the deprecated `networks` attribute

Junkern:
The differentiation between `networks` and `network_advanced` is a bit buggy, but that should suffice until we properly remove the `networks` with the next major version

PR #253, opened 2021-07-29
Add basic data pull for docker containers

stephenliberty:
Noticed a missing feature - the ability to use the `data` calls in terraform to pull information about a running container. This allows someone to query the docker API for container information, and potentially use it to trigger their own resource regeneration off of it.


stephenliberty:
@mavogel - another one that I'd like to ping you on. Can you do a review on this?

PR #266, opened 2021-08-10
chore(ci): adapt acc test docker version

mavogel:
follow up for #256 as we forgot to adapt those versions when updating the libs


suzuki-shunsuke (approved):
Thanks.

PR #305, opened 2021-11-02
Fix edge cases where tag might not be populated internally

abergmeier:
Better to look up by image id anyway since we just built this exact image.


suzuki-shunsuke (commented):
Thank you for your contribution.
Would you describe more detail?
What is solved? Why this is better?
If this is a bug fix, please create an issue with issue template. https://github.com/kreuzwerker/terraform-provider-docker/issues/new?assignees=&labels=&template=Bug_Report.md

abergmeier:
Am working on a provider based on podman.

Issue #81, opened 2020-12-25
Container is not stopped when using `rm=true`


mavogel:
<i>This issue was originally opened by @inercia as https://github.com/hashicorp/terraform-provider-docker/issues/163. It was migrated here as a result of the community provider takeover from @kreuzwerker. The original body of the issue is below.</i><hr>
### Terraform Version
Terraform v0.11.15-dev
+ provider.docker v2.0.0

### Affected Resource(s)
- docker_container

### Terraform Configuration Files
```hcl
provider "docker" {
  host = "tcp://127.0.0.1:2375/"
}

resource "docker_network" "network" {
  name            = "haproxy-test-net"
  check_duplicate = "true"
}

resource "docker_container" "haproxy-test" {
  name                  = "haproxy-test"
  image                 = "haproxy"
  start                 = true
  rm                    = true
  privileged            = true
  must_run              = true
  restart               = "no"
  destroy_grace_seconds = 60
  network_mode          = "bridge"

  upload {
    file    = "/usr/local/etc/haproxy/haproxy.cfg"
    content = <<EOF
    defaults
      timeout connect 10s
      timeout client 86400s
      timeout server 86400s

    frontend apiserver
      bind :6443
      default_backend apiserver-backend

    backend apiserver-backend
      option httpchk GET /healthz
      server localhost 127.0.0.1:6443 check check-ssl verify none
EOF

  }
}
```

### Debug Output
[cluster creation](https://gist.github.com/inercia/f1d83d434319d841a3a629c5e5b355cc)
[cluster destruction](https://gist.github.com/inercia/ce43d299af21623489655822e3818a1a)

### Expected Behavior
The docker container should have been properly stopped, but it is still running after `terraform destroy`. A `docker ps` shows the container running.

### Actual Behavior

From [the docs](https://www.terraform.io/docs/providers/docker/r/container.html#rm) _"If true, then the container will be automatically removed after his execution. Terraform won't check this container after creation."_, so it seems to be equivalent to the `--rm` flag of `docker`. However, the docker container is not stopped at all (let alone to be removed). Maybe the `rm` argument is not preoperly explained in the docs.

Looking at the code, it seems the only thing the provider does when destroying the container with `rm=true` [is to remove the ID](https://github.com/terraform-providers/terraform-provider-docker/blob/master/docker/resource_docker_container_funcs.go#L560-L563).

### Steps to Reproduce
1. Use `rm=true` in a container.
2. `terraform apply`
[...]

PR #321, opened 2021-12-02
Update registry_image.md

Skp2608:
Fixed #153  @mavogel Could you please check my pull request?

Thanks

PR #322, opened 2021-12-02
fix: pass container rm flag

mavogel:
- see #81
- fixed tests and tested with a [real build](https://github.com/kreuzwerker/terraform-provider-docker/blob/master/CONTRIBUTING.md#test-against-current-terraform-iac-descriptions) of the provider
- added `Stargazers over time`

### DIFF
--- a/docs/resources/registry_image.md
+++ b/docs/resources/registry_image.md
@@ -8,12 +8,9 @@ description: |-
 <!-- Bug: Type and Name are switched -->
 # docker_registry_image (Resource)
 
-Manages the lifecycle of docker image/tag in a registry means it can store one or more version
-of specific docker images and identified by their tags.
+Manages the lifecycle of docker image/tag in a registry.
 
 ## Example Usage
-
-To be able to update an image itself when an updated image arrives.
 
 ```terraform
 resource "docker_registry_image" "helloworld" {
--- a/docs/resources/registry_image.md
+++ b/docs/resources/registry_image.md
@@ -3,14 +3,16 @@
 page_title: "docker_registry_image Resource - terraform-provider-docker"
 subcategory: ""
 description: |-
-  Manages the lifecycle of docker image/tag in a registry.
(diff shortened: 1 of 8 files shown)


internal/provider/resource_docker_container_funcs.go
```diff
@@ -824,39 +824,47 @@ func resourceDockerContainerUpdate(ctx context.Context, d *schema.ResourceData,
 func resourceDockerContainerDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*ProviderConfig).DockerClient
 
-	if d.Get("rm").(bool) {
```
mavogel, line 827: I had no clue TBH why this was there, as it will never stop/remove the container if `rm=true` 🤔

docs/resources/registry_image.md
```diff
@@ -8,13 +8,10 @@ description: |-
 <!-- Bug: Type and Name are switched -->
 # docker_registry_image (Resource)
```
suzuki-shunsuke, line 10: https://github.com/kreuzwerker/terraform-provider-docker/pull/321 was reverted.

suzuki-shunsuke: How about updating Go code and generating the document again?

* https://github.com/kreuzwerker/terraform-provider-docker/blob/0da4706851ac8f86af531b5a1b725ae66e5b57af/internal/provider/resource_docker_registry_image.go#L11

mavogel: True, my bad. I missed this. I will update the go code and adapt to #321

mavogel: see https://github.com/kreuzwerker/terraform-provider-docker/pull/322/commits/5a0bbd78b60c64bf0878e927af88df4122e7038b @suzuki-shunsuke

suzuki-shunsuke (approved):

Issue #294, opened 2021-10-05
Support for prevent destruction of docker volumes


milost:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be very helpful if it was possible to prevent certain volumes from being destroyed during a `terraform destroy`. This way it would be possible, for example, to keep data persisted by a database during the development of the infrastructure (see configuration example below). For example, it would be possible to destroy the actual docker containers and images, but keep any data that is stored in volumes. At a later point in time, an infrastructure could be rebuilt in which, for example, a database has access to the data of a previously created volume.

### New or Affected Resource(s)

* docker_volume

### Potential Terraform Configuration

```hcl
terraform {
  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "~> 2.13.0"
    }
  }
}


resource "docker_image" "postgres" {
  name         = "postgres:${var.server_version}"
  keep_locally = false
}

resource "docker_volume" "postgres" {
  name = var.server_volume
  prevent_destruction = true  # prevent volume from being destroyed
}

resource "docker_container" "postgres" {
  image = docker_image.postgres.name
  name  = var.server_name
  env   = ["POSTGRES_USER=postgres", "POSTGRES_PASSWORD=12345"]
  ports {
    internal = 5432
    external = 5432
  }
  volumes {
    container_path = "/var/lib/postgresql/data"
    volume_name    = var.server_volume
  }
}

```

xadlien:
Would love to see this feature. Please don't close due to being stale. I'll try to make this change myself.

xadlien:
Please assign this to me. I have some working code and am trying to wrap a bow on the commits.

suzuki-shunsuke:
Thank you for your proposal.
I understand your motivation, but I wonder if we should really add such an attribute.

You can prevent the destruction of resources by running `terraform state rm` before `terraform destroy`.

```console
$ terraform state rm docker_volume.postgres
$ terraform destroy
```

Or

```console
$ terraform state list | grep -E "^docker_volume\." | xargs -n 1 terraform state rm
$ terraform destroy
```

[...]

PR #326, opened 2021-12-18
feat(docker_volume): add the prevent_destroy parameter to docker_volume (#294)

xadlien:
feat(docker_volume): add the prevent_destroy parameter to docker_volume (#294)

### DIFF
--- a/internal/provider/resource_docker_volume.go
+++ b/internal/provider/resource_docker_volume.go
@@ -136,7 +136,7 @@ func resourceDockerVolumeRead(ctx context.Context, d *schema.ResourceData, meta 
 
 func resourceDockerVolumeDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 
-	if preventDestroy := d.Get("prevent_destroy").(bool); preventDestroy {
+	if d.Get("prevent_destroy").(bool) {
 		return nil
 	}
 
--- a/internal/provider/resource_docker_volume.go
+++ b/internal/provider/resource_docker_volume.go
@@ -63,6 +63,12 @@ func resourceDockerVolume() *schema.Resource {
 				Description: "The mountpoint of the volume.",
 				Computed:    true,
 			},
+			"prevent_destroy": {
+				Type:        schema.TypeBool,
+				Description: "If true, then the Docker image won't be deleted on destroy operation. If this is false, it will delete the volume from the docker local storage on destroy operation.",
+				Optional:    true,
+				ForceNew:    true,
+			},
 		},
 		SchemaVersion: 1,
 		StateUpgraders: []schema.StateUpgrader{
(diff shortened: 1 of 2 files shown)


xadlien:
It's my first time contributing to a repo other than my own so I apologize if anything with this PR is off. Please let me know and I'll fix it.

internal/provider/resource_docker_volume.go
```diff
@@ -129,6 +135,11 @@ func resourceDockerVolumeRead(ctx context.Context, d *schema.ResourceData, meta
 }
 
 func resourceDockerVolumeDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+
+	if prevent_destroy := d.Get("prevent_destroy").(bool); prevent_destroy {
```
suzuki-shunsuke, line 139:
```suggestion
	if d.Get("prevent_destroy").(bool) {
```

In Go, you should use `camelCase` rather than `snake_case`.

internal/provider/resource_docker_volume.go
```diff
@@ -129,6 +135,11 @@ func resourceDockerVolumeRead(ctx context.Context, d *schema.ResourceData, meta
...
+		return nil
```
suzuki-shunsuke, line 140:
```suggestion
		d.SetId("")
		return nil
```

https://learn.hashicorp.com/tutorials/terraform/provider-setup#implement-read

I don't have a confidence, but maybe `d.SetId("")` is required.

Otherwise after you remove the resource definition,
`terraform plan` would include the destroy operation because the resource remains in the State.

suzuki-shunsuke:
Thank you for your contribution.

internal/provider/resource_docker_volume.go
xadlien: I updated and squashed into my original commit

internal/provider/resource_docker_volume.go
xadlien: I added `d.SetId("")` and got the same result. This works the same was as `keep_locally` for docker_images in the provider.

suzuki-shunsuke: Oh, probably I misunderstood.
`d.SetId("")` isn't needed.

xadlien:
Removed previous commit that was not needed.

internal/provider/resource_docker_volume.go
```diff
@@ -129,6 +135,11 @@ func resourceDockerVolumeRead(ctx context.Context, d *schema.ResourceData, meta
 }
 
 func resourceDockerVolumeDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+
+	if preventDestroy := d.Get("prevent_destroy").(bool); preventDestroy {
```
suzuki-shunsuke, line 139:
```suggestion
	if d.Get("prevent_destroy").(bool) {
```

suzuki-shunsuke:
https://github.com/kreuzwerker/terraform-provider-docker/issues/294#issuecomment-997383501

PR #363, opened 2022-03-24
docs(r/container): fix the example to keep container running

suzuki-shunsuke:


MrBenGriffin:
This does not resolve the problem - which is that both "must_run" and "start" are ignored despite being (by default) set to true - and even when explicitly set to true.

Meanwhile the command/tty lines are documented to do other things altogether, and this 'fix' is (AFAIK) effectively engineering a side effect of tty/command.

PR #369, opened 2022-04-18
Update the configs service example documentation

adriangabura:
The service docs clearly state that configs are a block.


Junkern:
@adriangabura thanks for the PR! The docs are auto-generated and the template for the specific part you are refering to can be found https://github.com/kreuzwerker/terraform-provider-docker/blob/master/examples/resources/docker_config/resource-no-downtime.tf

Would you mind updating the PR by modifying that file?

adriangabura:
> @adriangabura thanks for the PR! The docs are auto-generated and the template for the specific part you are refering to can be found https://github.com/kreuzwerker/terraform-provider-docker/blob/master/examples/resources/docker_config/resource-no-downtime.tf
> 
> Would you mind updating the PR by modifying that file?

Thank you! I updated the file in question!

Junkern:
@adriangabura could you fix the linting errors in the changed `resource-no-downtime.tf` file? https://github.com/kreuzwerker/terraform-provider-docker/runs/7555869544?check_suite_focus=true

PR #371, opened 2022-04-21
chore: remove the workflow to close stale issues and pull requests

### LINKED ISSUE #370: Abandon automatic close of stale issues and pull requests
We introduced the automatic close of stale issues and pull requests. #155 #157
But I think this automation doesn't fit our current situation.

This workflow would work well if we deal with issues and pull requests properly and only trivial issues and pull requests are stale.

But actually, many meaningful issues and pull requests are stale and closed automatically.
This isn't good experience.

So, I'd like to remove the workflow to close stale issues and pull requests.

https://github.com/kreuzwerker/terraform-provider-docker/blob/master/.github/workflows/stale.yaml

suzuki-shunsuke:
Close #370


suzuki-shunsuke:
CI failed, but it has nothing to do with this change.

mavogel (approved):
yes, let's remove it for now

Issue #280, opened 2021-09-02
docker_registry_image fails to exclude files listed in .dockerignore 


moritzheiber:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and docker Provider) Version

```console
terraform -v
Terraform v1.0.5
on linux_amd64
```

Provider version is `2.15.0`.

### Affected Resource(s)

* `docker_registry_image`

### Terraform Configuration Files

I've created a POC repository which allows for reproducing the bug: https://github.com/moritzheiber/terraform-docker-provider-dockerignore

### Expected Behaviour

The files added to the `.dockerignore` file are not contained within the resulting image.

### Actual Behaviour

The files added to the `.dockerignore` file are contained within the resulting image.

### Steps to Reproduce

See the `README.md` in the POC repository.

### References

- #240 

Also, taking a look at the test for this particular functionality, [the `Dockerfile`](https://github.com/kreuzwerker/terraform-provider-docker/blob/master/scripts/testing/docker_registry_image_context_dockerignore/Dockerfile) used for it doesn't have a broad instruction for adding files into the image, which means the ignored file cannot possible ever show up in the image itself, leading me to believe the test itself is not sufficient in order to verify the given functionality.

moritzheiber:
Uhm, this issue has been marked as a target for 2.16.0, it shouldn’t even be marked as stale..?

moritzheiber:
@mavogel some help? 😅

marcelmeulemans:
The issue here is that absolute paths are passed to the excludes matcher. I would create a PR but not being proficient in go is preventing me from fixing the associated unit test. The unit test currently does not really check the ignore behaviour because the file being modified in the test [here](https://github.com/kreuzwerker/terraform-provider-docker/blob/master/internal/provider/resource_docker_registry_image_test.go#L182) was not even copied to the image.

mavogel:
sure @moritzheiber feel free to help :)

PR #397, opened 2022-07-07
fix: compare relative paths when excluding, fixes kreuzwerker#280

marcelmeulemans:
Finally found time to write a test for the fix I proposed earlier in #280.

This fix itself is a trivial change from trying to match absolute paths to matching relative paths.

The bulk of changes are related to fixing the ignore file acceptance test which didn't really test anything at all due to a number of flaws (which this PR fixes):
- Dockerfile was not setup to COPY the ignored file so the test was not detecting that the ignore functionality did not work (the ignored file was added to the build context tar, but not to the image).
- The test looked like it wanted to compare the digest of an image before and after changing an ignored file. However it was only testing for the existence of an image digest.
- The test steps were using the same resource so terraform only actually built the image once.

The test now works properly which can be verified by changing the ignore file in which case the test will fail.

### DIFF
--- a/internal/provider/resource_docker_registry_image_funcs.go
+++ b/internal/provider/resource_docker_registry_image_funcs.go
@@ -292,7 +292,7 @@ func buildDockerImageContextTar(buildContext string) (string, error) {
 
 	pm, err := fileutils.NewPatternMatcher(excludes)
 	if err != nil {
-		return "", fmt.Errorf("unable to create pattern matcher from .dockerignore exlcudes - %v", err.Error())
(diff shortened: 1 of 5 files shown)


Junkern:
Thanks a lot for your PR! I am currently in the progress in rewriting parts of the `docker_registry_image` to fix the various issues with this resource. Those changes probably will be incompatible with your changes or make them obsolete...
However, I will use the enhancements to the tests in order to also verify the rewriting :)

Junkern:
Changed my mind :D Merging your contribution first makes my life easier. Thank you!

Issue #239, opened 2021-06-28
Cannot authenticate with multiple docker registries


adam-lebon:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and docker Provider) Version

Terraform version:
```
Terraform v1.0.0
on darwin_amd64
+ provider registry.terraform.io/kreuzwerker/docker v2.13.0
```

Docker for Mac version:
```
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Build with BuildKit (Docker Inc., v0.5.1-docker)
  compose: Docker Compose (Docker Inc., 2.0.0-beta.1)
  scan: Docker Scan (Docker Inc., v0.8.0)

 Server Version: 20.10.6
```
### Affected Resource(s)

* The provider configuration
* `docker_image`

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    docker = {
      source = "kreuzwerker/docker"
      version = "2.13.0"
    }
  }
}

provider "docker" {
  registry_auth {
      address = "registry.internal.example"
      config_file = file("~/.docker/config.json")
  }

  registry_auth {
      address = "registry.gitlab.com"
      config_file = file("~/.docker/config.json")
  }
}

resource "docker_image" "nginx" {
  name = "nginx:latest"
}

```

### Debug Output

https://gist.github.com/adam-lebon/56444f77c781c44b88caafb01b78da27

### Expected Behaviour

[As descibe in the documentation](https://registry.terraform.io/providers/kreuzwerker/docker/latest/docs#registry-credentials), we should be able to configure authentication to multiple registries.

### Actual Behaviour

The 2.12.0 release added a new parameter `MaxItems: 1` to the registry_auth block which prevents the usage of multiple private registry.

This result in terraform returning a error when validating the provider configuration :

```bash
$ terraform apply

│ Error: Too many registry_auth blocks
│ 
│   on  line 0:
│   (source code not available)
│ 
│ No more than 1 "registry_auth" blocks are allowed
```

### Steps to Reproduce

[...]

PR #400, opened 2022-07-11
fix: Enable authentication to multiple registries again.

Junkern:
Fixes https://github.com/kreuzwerker/terraform-provider-docker/issues/239

This is still a WIP until I have figured out how to have deal with the `ConflictsWith` (which is disabled for now). Opened an issue in the sdk repo: https://github.com/hashicorp/terraform-plugin-sdk/issues/1001


Junkern:
Bad news: We have to implement the ConflictsWith validation manually, as this is not supported by the `terraform-plugin-sdk`.

PR #404, opened 2022-07-13
Fix error handling in searchLoclImages()

prologic:
Hopefully this improves our ability to debug the masked error and potentially racey behaviour in infra that relies on the Docker provider via Terraform mover SSH.

cc @Junkern

### DIFF
--- a/internal/provider/resource_docker_image_funcs.go
+++ b/internal/provider/resource_docker_image_funcs.go
@@ -63,7 +63,11 @@ func resourceDockerImageRead(ctx context.Context, d *schema.ResourceData, meta i
 
 	imageName := d.Get("name").(string)
 
-	foundImage := searchLocalImages(ctx, client, data, imageName)
+	foundImage, err := searchLocalImages(ctx, client, data, imageName)
+	if err != nil {
+		return diag.Errorf("error looking up local image %q: %s", imageName, err)
+	}
+
 	if foundImage == nil {
 		log.Printf("[DEBUG] did not find image with name: %v", imageName)
 		d.SetId("")
@@ -106,20 +110,23 @@ func resourceDockerImageDelete(ctx context.Context, d *schema.ResourceData, meta
 }
 
 // Helpers
-func searchLocalImages(ctx context.Context, client *client.Client, data Data, imageName string) *types.ImageSummary {
+func searchLocalImages(ctx context.Context, client *client.Client, data Data, imageName string) (*types.ImageSummary, error) {
 	imageInspect, _, err := client.ImageInspectWithRaw(ctx, imageName)
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("unable to inspect imave %s: %w", imageName, err)
 	}
 
-	jsonObj, _ := json.MarshalIndent(imageInspect, "", "\t")
+	jsonObj, err := json.MarshalIndent(imageInspect, "", "\t")
+	if err != nil {
+		return nil, fmt.Errorf("error parsing inspect response: %w", err)
+	}
 	log.Printf("[DEBUG] Docker image inspect from readFunc: %s", jsonObj)
 
 	if apiImage, ok := data.DockerImages[imageInspect.ID]; ok {
 		log.Printf("[DEBUG] found local image via imageName: %v", imageName)
-		return apiImage
+		return apiImage, nil
 	}
-	return nil
+	return nil, nil
 }
 
 func removeImage(ctx context.Context, d *schema.ResourceData, client *client.Client) error {
@@ -138,7 +145,10 @@ func removeImage(ctx context.Context, d *schema.ResourceData, client *client.Cli
 		return fmt.Errorf("empty image name is not allowed")
 	}
 
-	foundImage := searchLocalImages(ctx, client, data, imageName)
+	foundImage, err := searchLocalImages(ctx, client, data, imageName)
+	if err != nil {
+		return fmt.Errorf("error looking up local image %q: %w", imageName, err)
+	}
 
 	if foundImage != nil {
 		imageDeleteResponseItems, err := client.ImageRemove(ctx, imageName, types.ImageRemoveOptions{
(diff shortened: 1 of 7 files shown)


internal/provider/resource_docker_network.go
```diff
@@ -155,51 +151,3 @@ func resourceDockerNetwork() *schema.Resource {
 		},
 	}
 }
-
```
Junkern, line 158: Could you re-add the deleted functions or is there a reason you deleted them?

internal/provider/resource_docker_network_test.go
```diff
@@ -325,48 +325,6 @@ func testAccNetworkIPv4(network *types.NetworkResource, internal bool) resource.
 	}
 }
 
-func TestAccDockerNetwork_ipv6(t *testing.T) {
```
Junkern, line 328: Same with the tests: Is there a reason you deleted them?

internal/provider/resource_docker_service_test.go
```diff
@@ -437,7 +437,7 @@ func TestAccDockerService_fullSpec(t *testing.T) {
 			s.Spec.TaskTemplate.ContainerSpec.Secrets[0].File.UID != "0" ||
 			s.Spec.TaskTemplate.ContainerSpec.Secrets[0].File.GID != "0" ||
 			// nolint: staticcheck
-			s.Spec.TaskTemplate.ContainerSpec.Secrets[0].File.Mode != os.FileMode(777) {
```
Junkern, line 440: I quickly looked into the code, and it seems like the whole parsing of the FileMode option is weird. I would ask you to leave this out of this PR and create an issue for that, so we can tackle this at a later point

internal/provider/resource_docker_image_funcs.go
```diff
@@ -106,20 +110,23 @@ func resourceDockerImageDelete(ctx context.Context, d *schema.ResourceData, meta
 }
 
 // Helpers
-func searchLocalImages(ctx context.Context, client *client.Client, data Data, imageName string) *types.ImageSummary {
+func searchLocalImages(ctx context.Context, client *client.Client, data Data, imageName string) (*types.ImageSummary, error) {
 	imageInspect, _, err := client.ImageInspectWithRaw(ctx, imageName)
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("unable to inspect imave %s: %w", imageName, err)
```
Junkern, line 116:
```suggestion
		return nil, fmt.Errorf("unable to inspect image %s: %w", imageName, err)
```

internal/provider/resource_docker_service_structures.go
```diff
@@ -1083,17 +1083,14 @@ func createServiceNetworks(v interface{}) ([]swarm.NetworkAttachmentConfig, erro
 func createLogDriver(v interface{}) (*swarm.Driver, error) {
 	logDriver := swarm.Driver{}
 	if len(v.([]interface{})) > 0 {
-		for _, rawLogging := range v.([]interface{}) {
```
Junkern, line 1086: Why did you remove the loop?

Junkern (changes requested):

internal/provider/resource_docker_network.go
prologic: Ahh yes sorry I can. I deleted it because of my own OCD when the linter yells at my for "unused" functions 😅

prologic: Done!

internal/provider/resource_docker_service_test.go
prologic: Okay! One sec...

prologic: Fix removed from this patch

internal/provider/resource_docker_service_structures.go
prologic: If you have a look at the TODO(s) there and the logic, I _think_ it only goes thorugh the loop once anyway, so the loop is superfluous. Linter also says so 😅

Junkern (changes requested):
One small change. Afterwards we can merge (if all test pass)

internal/provider/resource_docker_service_structures.go
prologic: Can you link me to where you found this `MAX_ELEMENTS: 1`? I'm getting a bit lost here in the code, api docs and what not 😅

Junkern: https://github.com/kreuzwerker/terraform-provider-docker/blob/master/internal/provider/resource_docker_service.go#L687

prologic:
> One small change. Afterwards we can merge (if all test pass). I have the feeling by changing the error handling we might have to adapt a few tests as well, but lets see. (You can also run them locally)

I _might_ need your help here,  I'm not actually sure why the tests pass for some versions of Go but fails for others.

I've ran `TestAccDockerService_convergeAndStopGracefully` locally here on my workstation and it passes just fine. Hmmm ? 🤔

I _hope_ I haven't changed the behaviour of anything!

Junkern:
We did indeed change the behaviour. Before, if the `ImageInspectWithRaw` call within the `searchLocalImages` did not find any image, it simply did not error, and the parentFunction would then continue searching for the image in the registry.

Now the implementation fails as soon as an image cannot be found locally. But sometimes the image is not stored locally. I tested the following implementation and it seems to work for a few selected tests (I cannot push to your branch):
```go
func searchLocalImages(ctx context.Context, client *client.Client, data Data, imageName string) (*types.ImageSummary, error) {
	imageInspect, _, err := client.ImageInspectWithRaw(ctx, imageName)
	if err != nil {
	        // the if below is new
		if errdefs.IsNotFound(err) {
			return nil, nil
		}
		return nil, fmt.Errorf("unable to inspect image %s: %w", imageName, err)
	}
```

And you also have to `go get github.com/docker/docker/errdefs` to include the correct package

Junkern:
I did include the `searchLocalImage` changes in the newest `v2.18.1` version (which I really wanted to release today).
The changes regarding the `internal/provider/resource_docker_service_structures.go` file will be part of one of the next releases :)

Thanks for contributing! I hope the better error handling helps you

prologic:
@Junkern Thank you so much! Sorry I didn't quite get this over the line, yesterday for me was. nuts, over 6hrs of meetings, didn't get any "real work" done 😅

PR #427, opened 2022-08-09
fix: Check the operating system for determining the default Docker socket

kylecarbs:
Windows doesn't use `/var/run/docker.sock`, so this provider
didn't work out of the box.


Junkern (approved):

internal/provider/provider.go
```diff
@@ -39,9 +40,17 @@ func New(version string) func() *schema.Provider {
...
-					Required:    true,
-					DefaultFunc: schema.EnvDefaultFunc("DOCKER_HOST", "unix:///var/run/docker.sock"),
+					Type:     schema.TypeString,
+					Required: true,
+					DefaultFunc: func() (interface{}, error) {
+						if v := os.Getenv("DOCKER_HOST"); v != "" {
+							return v, nil
+						}
+						if runtime.GOOS == "windows" {
+							return "npipe:////.//pipe//docker_engine", nil
```
Junkern, line 50: Asking because I don't own a windows: I double-checked with https://docs.docker.com/desktop/faqs/general/ and there the value is `npipe:////./pipe/docker_engine`, only one `/` between `pipe` and `docker_engine`.
Which is the correct value?

Junkern (changes requested):

internal/provider/provider.go
kylecarbs: Fixed! It seems to work with both.

Junkern (approved):

kylecarbs:
Thanks, @Junkern! Feel free to merge :partying_face: 

This will help make our experience much smoother, so it'd be amazing if we could get a new release soon!

Junkern:
@kylecarbs released with `v1.20.2`

kylecarbs:
You rock @Junkern :star_struck:

PR #438, opened 2022-08-19
Adding Support for Windows Paths in Bash

madduci:
Included support for Absolute Paths in Windows that begin with e.g. `G:/`

Tests performed also with an online RegEx checker: regexr.com/6s741


Junkern:
Thanks for your contribution!

PR #451, opened 2022-09-02
fix: oauth authorization support for azurecr

wSedlacek:
This adds support for authorization with Azure Container Registry.
Previously this would fail for 2 reasons:
1. It was trying to use basic for azurecr
2. It was not looking for the `access_token` in the oauth response

### DIFF
--- a/internal/provider/resource_docker_registry_image_funcs.go
+++ b/internal/provider/resource_docker_registry_image_funcs.go
@@ -14,7 +14,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -483,7 +482,16 @@ func deleteDockerRegistryImage(pushOpts internalPushImageOptions, registryWithPr
 	}
 
 	if username != "" {
-		req.SetBasicAuth(username, password)
+		if pushOpts.Registry != "ghcr.io" && !isECRRepositoryURL(pushOpts.Registry) && !isAzureCRRepositoryURL(pushOpts.Registry) && pushOpts.Registry != "gcr.io" {
+			req.SetBasicAuth(username, password)
+		} else {
+			if isECRRepositoryURL(pushOpts.Registry) {
+				password = normalizeECRPasswordForHTTPUsage(password)
+				req.Header.Add("Authorization", "Basic "+password)
+			} else {
+				req.Header.Add("Authorization", "Bearer "+base64.StdEncoding.EncodeToString([]byte(password)))
+			}
+		}
 	}
 
 	setupHTTPHeadersForRegistryRequests(req, fallback)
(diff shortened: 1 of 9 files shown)


wSedlacek:
Still gets an error when deleting, but creating works now.

> Got error deleting registry image: Bad credentials: 401 Unauthorized

internal/provider/resource_docker_registry_image_funcs.go
```diff
@@ -500,56 +508,26 @@ func deleteDockerRegistryImage(pushOpts internalPushImageOptions, registryWithPr
...
-			oauthResp, err := client.Do(req)
-			if err != nil {
-				return err
-			}
-			switch oauthResp.StatusCode {
-			case http.StatusOK, http.StatusAccepted, http.StatusNotFound:
-				return nil
-			default:
-				return fmt.Errorf("Got bad response from registry: " + resp.Status)
-			}
```
wSedlacek, line 547: This logic looked to be the same as what getImageDigest was doing for `getAuthToken()` so I used that function instead and it seems to work.

wSedlacek:
@Junkern These changes are working in my local test using Azure. My organization is blocked by the issues that this fixes.
Would love to get a review on this.

internal/provider/resource_docker_registry_image_funcs.go
```diff
@@ -483,7 +482,16 @@ func deleteDockerRegistryImage(pushOpts internalPushImageOptions, registryWithPr
...
+		if pushOpts.Registry != "ghcr.io" && !isECRRepositoryURL(pushOpts.Registry) && !isAzureCRRepositoryURL(pushOpts.Registry) && pushOpts.Registry != "gcr.io" {
+			req.SetBasicAuth(username, password)
+		} else {
+			if isECRRepositoryURL(pushOpts.Registry) {
+				password = normalizeECRPasswordForHTTPUsage(password)
+				req.Header.Add("Authorization", "Basic "+password)
+			} else {
+				req.Header.Add("Authorization", "Bearer "+base64.StdEncoding.EncodeToString([]byte(password)))
+			}
+		}
```
Junkern, line 494: I am unsure about this addition.
When I tested the different registries a few weeks ago, for the `DELETE` route the existing implementation, using basicAuth was sufficient.
I will test this again

internal/provider/resource_docker_registry_image_funcs.go
Junkern: Nice!

Junkern (commented):
I have a local test suite which tests against many registries. I will have to add an Azure registry and rerun the test-suite to make sure that everything still works the same. This could take at least a few days.

internal/provider/resource_docker_registry_image_funcs.go
wSedlacek: I was very specifically getting this error without these changes.
```
Got error deleting registry image: Bad credentials: 401 Unauthorized
```

wSedlacek:
Just noticed the formatting errors. Fixing now.

Junkern (approved):

Junkern:
I'll merge this and do some more testing with my local registry test stack.

PR #454, opened 2022-09-07
Configurable timeout for docker_container resource stateChangeConf

sullivtr:
Hi,
The docker_container resource provides a hardcoded timeout for the terraform to wait for the container to be ready. However, it makes sense to have this value be configurable to allow for slower starting containers. The default will remain `15` seconds, with an option to override on a container by container basis.

### DIFF
--- a/internal/provider/resource_docker_container.go
+++ b/internal/provider/resource_docker_container.go
@@ -926,6 +926,13 @@ func resourceDockerContainer() *schema.Resource {
 				},
 			},
 
+			"container_read_refresh_timeout": {
+				Type:        schema.TypeInt,
+				Description: "The total number of seconds to wait for the container to reach status 'running'",
+				Optional:    true,
+				Default:     15,
+			},
+
 			"sysctls": {
 				Type:        schema.TypeMap,
 				Description: "A map of kernel parameters (sysctls) to set in the container.",
--- a/internal/provider/resource_docker_container_funcs.go
+++ b/internal/provider/resource_docker_container_funcs.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	containerReadRefreshTimeout             = 15 * time.Second
+	// containerReadRefreshTimeout             = 15 * time.Second
 	containerReadRefreshWaitBeforeRefreshes = 100 * time.Millisecond
 	containerReadRefreshDelay               = 100 * time.Millisecond
 )
@@ -560,6 +560,7 @@ func resourceDockerContainerCreate(ctx context.Context, d *schema.ResourceData, 
 }
 
 func resourceDockerContainerRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	containerReadRefreshTimeout := d.Get("container_read_refresh_timeout").(int)
 	log.Printf("[INFO] Waiting for container: '%s' to run: max '%v seconds'", d.Id(), containerReadRefreshTimeout)
 	client := meta.(*ProviderConfig).DockerClient
 
@@ -577,7 +578,7 @@ func resourceDockerContainerRead(ctx context.Context, d *schema.ResourceData, me
 		Pending:    []string{"pending"},
 		Target:     []string{"running"},
 		Refresh:    resourceDockerContainerReadRefreshFunc(ctx, d, meta),
-		Timeout:    containerReadRefreshTimeout,
+		Timeout:    time.Duration(containerReadRefreshTimeout) * time.Second,
 		MinTimeout: containerReadRefreshWaitBeforeRefreshes,
 		Delay:      containerReadRefreshDelay,
 	}
--- a/internal/provider/resource_docker_container_funcs.go
+++ b/internal/provider/resource_docker_container_funcs.go
@@ -29,7 +29,6 @@ import (
 )
 
 const (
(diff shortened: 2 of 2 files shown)


internal/provider/resource_docker_container.go
```diff
@@ -926,6 +926,13 @@ func resourceDockerContainer() *schema.Resource {
 				},
 			},
 
+			"container_read_refresh_timeout_milliseconds": {
+				Type:        schema.TypeInt,
+				Description: "The total number of seconds to wait for the container to reach status 'running'",
```
asan127, line 931:
```suggestion
				Description: "The total number of milliseconds to wait for the container to reach status 'running'",
```

sullivtr: Thanks fixed

internal/provider/resource_docker_container_funcs.go
```diff
@@ -560,7 +559,8 @@ func resourceDockerContainerCreate(ctx context.Context, d *schema.ResourceData,
 }
 
 func resourceDockerContainerRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	log.Printf("[INFO] Waiting for container: '%s' to run: max '%v seconds'", d.Id(), containerReadRefreshTimeout)
+	containerReadRefreshTimeoutMilliseconds := d.Get("container_read_refresh_timeout_milliseconds").(int)
+	log.Printf("[INFO] Waiting for container: '%s' to run: max '%v seconds'", d.Id(), containerReadRefreshTimeoutMilliseconds)
```
Junkern, line 563:
```suggestion
	log.Printf("[INFO] Waiting for container: '%s' to run: max '%v seconds'", d.Id(), containerReadRefreshTimeoutMilliseconds/1000)
```

I think it makes sense to show the correct seconds, instead of milliseconds. My suggestion is probably not the best one, but you get the idea

Junkern (changes requested):

internal/provider/resource_docker_container_funcs.go
sullivtr: Thanks. I changed this over to milliseconds last minute, and missed this fmt "seconds" left behind. I will update.

sullivtr: Fixed this, and also ran the website generation. Still trying to figure out why those tests are failing (they timeout waiting for status running, so my configuration is wrong somewhere, as it should be reading the default)

sullivtr:
> 

Yes. I just fixed the doc generation, and I am working on the test now. I was relying on the automation to run the tests in Gitlab originally, although I didnt think they would fail given how they are currently set up.

sullivtr:
The basic, and init test seem to not be getting hydrated with the new config, but all of the other tests do. Not sure why yet, but I am investigating. Will update sometime today.

internal/provider/resource_docker_container_funcs.go
```diff
@@ -560,7 +560,13 @@ func resourceDockerContainerCreate(ctx context.Context, d *schema.ResourceData,
 }
 
 func resourceDockerContainerRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	log.Printf("[INFO] Waiting for container: '%s' to run: max '%v seconds'", d.Id(), containerReadRefreshTimeout)
+	containerReadRefreshTimeoutMilliseconds := d.Get("container_read_refresh_timeout_milliseconds").(int)
+	// Ensure the timeout can ever be 0, the default integer value.
```
Junkern, line 564:
```suggestion
	// Ensure the timeout can never be 0, the default integer value.
```

internal/provider/resource_docker_container_funcs.go
```diff
@@ -29,9 +29,9 @@ import (
 )
 
 const (
-	containerReadRefreshTimeout             = 15 * time.Second
-	containerReadRefreshWaitBeforeRefreshes = 100 * time.Millisecond
-	containerReadRefreshDelay               = 100 * time.Millisecond
+	containerReadRefreshTimeoutMillisecondsDefault = 15000
```
Junkern, line 32:
```suggestion
	containerReadRefreshTimeoutMillisecondsDefault = 15000 * time.Millisecond
```
Maybe we should be consistent here?

Junkern (changes requested):
Just two really small comments. In general I share your opinion about the issue of importang an container, but I also don't really know a better way for now.

internal/provider/resource_docker_container_funcs.go
sullivtr: So I thought about this, but I also use this for the default value set in the provider schema as well (just to avoid having two separate hardcoded integers of the same value) --> and the schema itself wouldn't accept a duration as an input type. Im happy to do whatever you think is best here tho.

internal/provider/resource_docker_container_funcs.go
sullivtr: thank you, fixed

Junkern (approved):

sullivtr:
@Junkern Is there a timeline for when I can expect this to be released? I know the contrib.md documents the release process, but I cant run that from my fork, 

Thank you!

Junkern:
Just triggered the release for `v2.22.0`
Should be available in a few minutes

PR #462, opened 2022-09-21
fix: Correct provider name to match the public registry

benley:
This fixes the following error that shows up if you run the provider in debug mode:

```json
{
  "@level": "info",
  "@message": "2022/09/21 15:25:08 [ERROR] Error parsing provider name \"registry.terraform.io/kreuzwerker/terraform-provider-docker\": Invalid provider type: Provider source \"kreuzwerker/terraform-provider-docker\" has a type with the prefix \"terraform-provider-\", which isn't valid. Although that prefix is often used in the names of version control repositories for Terraform providers, provider source strings should not include it.\n\nDid you mean \"kreuzwerker/docker\"?",
  "@timestamp": "2022-09-21T15:25:08.353718-04:00"
}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,7 +35,7 @@ func main() {
 	opts := &plugin.ServeOpts{ProviderFunc: provider.New(version)}
 
 	if debugMode {
-		debugOpts := &plugin.ServeOpts{ProviderFunc: provider.New(version), ProviderAddr: "registry.terraform.io/kreuzwerker/terraform-provider-docker", Debug: true}
+		debugOpts := &plugin.ServeOpts{ProviderFunc: provider.New(version), ProviderAddr: "registry.terraform.io/kreuzwerker/docker", Debug: true}
 		plugin.Serve(debugOpts)
 		return
 	}


Junkern (changes requested):
https://registry.terraform.io/kreuzwerker/docker leads to a 404 for me, https://registry.terraform.io/providers/kreuzwerker/docker does resolve. Could you change that?

benley:
> 

Oddly enough it seems to insist on the form that leads to a 404 when loaded as a URL, because I see another error when I test with `providers/` in the provider name:

```json
{
  "@level": "info",
  "@message": "2022/09/22 12:49:55 [ERROR] Error parsing provider name \"registry.terraform.io/providers/kreuzwerker/docker\": Invalid provider source string: The \"source\" attribute must be in the format \"[hostname/][namespace/]name\"","@timestamp":"2022-09-22T12:49:55.016795-04:00"
}
```

So I believe the string used in this PR is actually correct.

Junkern:
Tests are failing due to new image sha of busybox image. WIll fix later

benley:
Thanks!

Issue #481, opened 2022-11-18
Crash During Destroy


cozzbie:
Version:
```
Terraform v1.3.5
on darwin_amd64
```

When using the `docker_registry_image`, I can `apply` successfully,  pushing an image to gcr. However, during destroy, this happens

```
│ Error: Request cancelled
│
│ The plugin.(*GRPCProvider).ApplyResourceChange request was cancelled.
╵
╷
│ Error: Request cancelled
│
│ The plugin.(*GRPCProvider).ApplyResourceChange request was cancelled.
╵

Stack trace from the terraform-provider-docker_v2.23.0 plugin:

panic: runtime error: index out of range [1] with length 1

goroutine 56 [running]:
github.com/terraform-providers/terraform-provider-docker/internal/provider.parseAuthHeader({0xc000544001?, 0x1272101?})
	github.com/terraform-providers/terraform-provider-docker/internal/provider/data_source_docker_registry_image.go:162 +0x1a5
github.com/terraform-providers/terraform-provider-docker/internal/provider.getAuthToken({0xc000544001?, 0x1a5eadc?}, {0xc000036410, 0xf}, {0xc0002e4000, 0xdd}, 0x0?)
	github.com/terraform-providers/terraform-provider-docker/internal/provider/data_source_docker_registry_image.go:185 +0x5e
github.com/terraform-providers/terraform-provider-docker/internal/provider.deleteDockerRegistryImage({{0xc00059a640, 0x3f}, {0xc00045d6c0, 0x3f}, {0xc00059a640, 0x6}, {0xc0006b6000, 0xe}, {0xc0007f4000, 0x2f}, ...}, ...)
	github.com/terraform-providers/terraform-provider-docker/internal/provider/resource_docker_registry_image_funcs.go:515 +0x78f
github.com/terraform-providers/terraform-provider-docker/internal/provider.resourceDockerRegistryImageDelete({0x1bbb468?, 0xc0002dad50?}, 0x0?, {0x18e1a00?, 0xc000030770})
	github.com/terraform-providers/terraform-provider-docker/internal/provider/resource_docker_registry_image_funcs.go:99 +0x2c9
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).delete(0xc0002728c0, {0x1bbb468, 0xc0002dad50}, 0xd?, {0x18e1a00, 0xc000030770})
[...]

PR #482, opened 2022-11-19
fix: Handle Auth Header Scopes

cozzbie:
As reported in #481, there is an issue with handling auth headers with more than one scope. For example

```
header := "Bearer realm="https://gcr.io/v2/token",service="gcr.io",scope="repository:<owner>/:<repo>/<name>:pull””
```
Works fine with single scopes. However, once we have more than one scope like

```
header := "Bearer realm="https://gcr.io/v2/token",service="gcr.io",scope="repository:<owner>/:<repo>/<name>:push,pull""
```

https://github.com/kreuzwerker/terraform-provider-docker/blob/ffb0cbfb75f5fc331777e33c6eab6c476b9e88cf/internal/provider/data_source_docker_registry_image.go#L156

Essentially returns a trailing map value of ["pull"] which breaks on https://github.com/kreuzwerker/terraform-provider-docker/blob/ffb0cbfb75f5fc331777e33c6eab6c476b9e88cf/internal/provider/data_source_docker_registry_image.go#L162

So this yields something like...
```
realm="https://gcr.io/v2/token"
service="gcr.io"
scope="repository:<owner>/:<repo>/<name>:push"
pull
```
Which causes the error in #481.

This fix ensures that trailing scopes do not cause issues. So now we have...

```
realm="https://gcr.io/v2/token"
service="gcr.io"
scope="repository:<owner>/:<repo>/<name>:push,pull"
```

### DIFF
--- a/internal/provider/data_source_docker_registry_image.go
+++ b/internal/provider/data_source_docker_registry_image.go
@@ -158,9 +158,12 @@ func parseAuthHeader(header string) map[string]string {
 
 	for _, part := range parts {
 		vals := strings.SplitN(part, "=", 2)
-		key := vals[0]
-		val := strings.Trim(vals[1], "\", ")
-		opts[key] = val
+
+		if len(vals) == 2 {
+			key := vals[0]
+			val := strings.Trim(vals[1], "\", ")
+			opts[key] = val
+		}
 	}
 
 	return opts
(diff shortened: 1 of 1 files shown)


cozzbie:
@suzuki-shunsuke can I get an amen?

Junkern:
Thanks for the contribution, looks good! The two failing tests are due to changes in shasums of docker tags, has nothing to do with your changes.

cozzbie:
Thanks @Junkern

PR #541, opened 2023-04-18
docs: clarify availability of docker_image.repo_digest

siddharthab:
The detailed note could save users some troubleshooting time because without this knowledge, when building an image, it's confusing why the value is missing when first time a config is applied, but present on subsequent times.

The unit tests do not check for `repo_digest` for configurations which build an image.


Junkern:

PR #546, opened 2023-04-27
Update glob pattern in image docs

akeamc:
The glob pattern specified in the current docs (`src/*`) does not work on my machine, as it only matches top-level files in `src`. `src/**` matches files in all subdirectories, which I suspect is the indented behavior.


Junkern:

Issue #563, opened 2023-06-12
Can't set size_bytes in tmpfs_options on docker_service mount


JanKoppe:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and docker Provider) Version

Terraform v1.4.6
on linux_amd64
+ provider registry.terraform.io/kreuzwerker/docker v3.0.2

### Affected Resource(s)

* `docker_service`

### Terraform Configuration Files

```hcl
terraform {
  required_version = "~>1.0"

  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "~>3.0"
    }
  }
}

provider "docker" {}

resource "docker_service" "primary" {
  name = "foobar"
  task_spec {
    container_spec {
      image = "busybox"
      mounts {
        target = "/dev/shm"
        type   = "tmpfs"
        tmpfs_options {
          mode       = 644
          size_bytes = 256 * 1024 * 1024 * 1024 # 256 MBytes
        }
      }
    }
  }
}
```

### Debug Output
https://gist.github.com/JanKoppe/d93bf3ef99103f7b76e3461f8d1260ae
### Expected Behaviour

Docker service with the requested tmpfs mount should have been created

### Actual Behaviour

Terraform will finish the plan step, but crash in the apply step. No resources are created.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

A quick glance at the code shows a discrepancy between data types used here:

https://github.com/kreuzwerker/terraform-provider-docker/blob/master/internal/provider/resource_docker_service_structures.go#L835

and here: 

https://github.com/kreuzwerker/terraform-provider-docker/blob/master/internal/provider/resource_docker_service_structures.go#L256

I believe the later one should be changed to int64 as well. I'd like to offer trying to work on a fix for this.

### References

* #0000

JanKoppe:
Started working on this, btw.

JanKoppe:
First possible fix in PR, it's not a perfect solution, but the best that I can muster up with my limited Go skills.

PR #570, opened 2023-07-07
fix: resource_docker_service tmpfs mounts #563

JanKoppe:
terraform schemas can only work with int type which is either int32 or int64 equivalent depending on architecture, but the Docker lib expects an int64 for the size_bytes. Previously, the code assumed that the terraform schema would return an int64, which lead to provider crashes.

values are now being converted between int64 and int to handle this. The downside is possible value truncation: resources managed by Terraform can't have tmpfs size_bytes greater than just under 2 GiB (2147483647 byes by golang spec), which can lead to false information if the physical service has been modified to a tmfs size_bytes greater than that outside of terraform, because the int64->int conversion will possibly truncate that on go implementations where int is equivalent to int32.

As long as the limitiation of only TypeInt being available in the schema (and not e.g. TypeInt64) there will be no clean solution to this. An improvement could be to switch to either TypeFloat (53 bits of accuracy, still with the issue, but much less likely to impact in real life, it's 8TiB) or TypeString and convert the numbers between string representations. Both options exceed my Go capabilities though, I think.

### DIFF
--- a/internal/provider/resource_docker_service_structures.go
+++ b/internal/provider/resource_docker_service_structures.go
@@ -254,7 +254,7 @@ func flattenServiceMounts(in []mount.Mount) *schema.Set {
 			tmpfsOptionsItem := make(map[string]interface{})
 
 			tmpfsOptionsItem["size_bytes"] = int(v.TmpfsOptions.SizeBytes)
-			tmpfsOptionsItem["mode"] = v.TmpfsOptions.Mode.Perm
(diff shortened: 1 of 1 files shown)


wojciechorzechowskikreuzwerker (approved):

tinexw:
Hi @JanKoppe, thanks for the fix! Would you mind adding a unit test to
- document the truncation issue you mentioned above 
- (if possible) one that demonstrates the issue and the fix

Issue #331, opened 2021-12-29
Support for cpu limits


danwestness:
### Community Note
* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
I would like to be able to define "--cpus" via the terraform provider to more effectively profile resource contraints among containers

### New or Affected Resource(s)
docker_container

### Potential Terraform Configuration

resource "docker_container" "container1" {
  name     = "container1"
  image    = container/container
  cpus      =  "1.5"
}

### References

https://docs.docker.com/config/containers/resource_constraints/#cpu

mc-marcy:
maybe we can reopen this request. was also searching for this :)

PR #575, opened 2023-07-31
feat: support setting cpu shares

Blackskyliner:
This PR implements #331

### DIFF
--- a/docs/resources/container.md
+++ b/docs/resources/container.md
@@ -50,6 +50,7 @@ resource "docker_image" "ubuntu" {
 - `domainname` (String) Domain name of the container.
 - `entrypoint` (List of String) The command to use as the Entrypoint for the container. The Entrypoint allows you to configure a container to run as an executable. For example, to run `/usr/bin/myprogram` when starting a container, set the entrypoint to be `"/usr/bin/myprogra"]`.
 - `env` (Set of String) Environment variables to set in the form of `KEY=VALUE`, e.g. `DEBUG=0`
+- `cpus` (String) Specify how much of the available CPU resources a container can use. e.g a value of 1.5 means the container is guaranteed at most one and a half of the CPUs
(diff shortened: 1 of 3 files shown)


jizusun:
Hi @Blackskyliner , thank you for such a great work!

I'm wondering how can I use your fork before it's merged?

Blackskyliner:
@jizusun 

As far as I know you will have to follow this: https://developer.hashicorp.com/terraform/language/providers/requirements#in-house-providers

So register it with either a private registry or under your own namespace in the official registry.

twsl:
would love to see this merged

Blackskyliner:
I don't get why this is not moving anywhere. 
No initial review done in almost a year now. 

There is need for this by some, it makes sense to integrate it as it was done for gpus. 

So is this whole repository/plugin deprecated not maintained anymore?

I can also see that its not top-priority to look into this repository until they need it for their business but it would be great if at least on an occasional basis one of the authorized people could look into some PRs. Otherwise the PR feature could just be disabled to indicate that nothing can be contributed because of never-reviewed-because-of-time-or-incentive-constraints, as contributors also take their time to provide the patch as PR which may also lead to double implementations as seen here where now both implementations will need to be compared as which to integrate if a review gets to get going, increasing the workload for the reviewer.

enc:
We are moving forward to fix the pipeline in the newest PR. If this is done, we will work on all the PRs. Unfortunately it will take a short while to get up to speed in the project. Your support would be apprechiated. #613

Blackskyliner:
Rebased onto current master w/o problems and/or any conflicts. 

Do I need to add a test for the changes or do we assume that the config parsing and used library is well-tested enough.

Blackskyliner:
@enc I can upgrade my PR to latest master but would only do that if it would get merged in a timely manner.

Junkern (approved):

PR #608, opened 2024-03-11
feat: add parameters for cpu quota handling

janLo:
This exposes the options cpu_period and cpu_quota to the terraform docker_container resource and adds the convenience parameter `cpus` as it is used by the docker cli.

See https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler for upstream details


janLo:
I've just realized, that #575 does a similar thing.

enc:
Do you like in integrate the changes from #575 I think this would be great and I would love to merge it.

janLo:
I'm happy to make changes but I think I don't completely get what you meant.

matifali:
Hi @Junkern can we get this in too?

PR #609, opened 2024-03-12
feat: allow to set the cgroup parent for container

janLo:
It is already possible to set this to the image builder but not to the container.


enc (approved):
LGTM

janLo:
Can we get this moving somehow? We really need this and have to resort to a patched version at the moment.

Junkern (approved):

PR #613, opened 2024-04-26
chore: update Go version to 1.22 for consistency across workflows, jo…

enc:
…bs, and modules.


.go-version
```diff
@@ -1 +1 @@
-1.18
+1.22
```
salehparsa, line 1: I'm not sure but It might be make sense to update documentation [here](https://github.com/kreuzwerker/terraform-provider-docker/blob/master/CONTRIBUTING.md#set-up-your-machine) as well and state that the required version is 1.22+ instead of 1.18+

PR #641, opened 2024-08-16
fix(docker_container): only wait when healthcheck defined

### LINKED ISSUE #569: Plugin crashes when using `wait=true` with no healthcheck command
### Community Note

### Terraform (and docker Provider) Version

Terraform v1.5.2
on darwin_arm64

### Affected Resource(s)

* `docker_container`

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "3.0.2"
    }
  }
}

resource "docker_image" "nginx" {
  name         = "nginx:latest"
  keep_locally = true
}

resource "docker_container" "nginx" {
  image = docker_image.nginx.name
  name  = "nginx"
  ports {
    internal = 80
    external = 8080
  }
  wait = true
}
```

### Panic Output

```
Stack trace from the terraform-provider-docker_v3.0.2 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10317c3f4]

goroutine 97 [running]:
github.com/terraform-providers/terraform-provider-docker/internal/provider.resourceDockerContainerCreate.func1(0x0?)
        github.com/terraform-providers/terraform-provider-docker/internal/provider/resource_docker_container_funcs.go:508 +0x154
created by github.com/terraform-providers/terraform-provider-docker/internal/provider.resourceDockerContainerCreate
        github.com/terraform-providers/terraform-provider-docker/internal/provider/resource_docker_container_funcs.go:521 +0x2384

Error: The terraform-provider-docker_v3.0.2 plugin crashed!
```

### Expected Behaviour

Since its `wait` value is `true`, it should wait until healthy. But it does not have `healthcheck` value, so there is no
way to check it. So, it should raise error for `terraform apply` or `terraform validate` commands.

### Actual Behaviour

1. It doesn't raise any error for `terraform validate` command. 
2. It crashes when executing `terraform apply`.

### Steps to Reproduce

1. `terraform apply`

FalcoSuessgott:
fixes #569 

perhaps it makes sense to move `wait` and `wait_timeout` attribute into `healthcheck` block, to make the corelation more obvious?

### DIFF
--- a/internal/provider/resource_docker_container_funcs.go
+++ b/internal/provider/resource_docker_container_funcs.go
@@ -496,28 +496,40 @@ func resourceDockerContainerCreate(ctx context.Context, d *schema.ResourceData, 
 			return diag.Errorf("Unable to start container: %s", err)
 		}
 
-		if d.Get("wait").(bool) {
+
+		_ , hasHealthCheck := d.GetOk("healthcheck")
+
+		if d.Get("wait").(bool) && hasHealthCheck {
 			waitForHealthyState := func(result chan<- error) {
 				for {
 					infos, err := client.ContainerInspect(ctx, retContainer.ID)
 					if err != nil {
 						result <- fmt.Errorf("error inspecting container state: %s", err)
-					}
-					//infos.ContainerJSONBase.State.Health is only set when there is a healthcheck defined on the container resource
-					if infos.ContainerJSONBase.State.Health.Status == types.Healthy {
-						log.Printf("[DEBUG] container state is healthy")
 						break
 					}
+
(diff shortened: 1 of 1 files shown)


Junkern:
Thanks for the PR! On the first glance this seems correct - but what about containers which have a built-in health-check? For those containers I do not have to define a `health_check` block myself. And then the `wait` would not perform as I expect. 

Maybe we should check if the deploy container has a healthcheck (either from the health check block or via a `HEALTHCHECK` directive from the container). If both things are not true, then we can abort. I will work on a test case, so that we can finally fix #569

Junkern:
Implemented by #709

Issue #642, opened 2024-08-24
Docker Container List data source


KetchupBomb:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Looking for a data source for listing containers. Specifically, [the `/containers` Docker Engine API endpoint](https://docs.docker.com/reference/api/engine/v1.45/#tag/Container/operation/ContainerList). 

### New or Affected Resource(s)

* New `data` `docker_containers`

### Potential Terraform Configuration

```hcl
data "docker_containers" "this" {}
ouput "container_fingerprint" {
  description = "Toy example mapping Container IDs to an md5sum fingerprint of its configuration"
  value       = { for c in data.docker_containers.this : c.ID => md5(jsonencode(c)) }
}
```

### References

Thanks for this Provider!

* [Docker Engine API endpoint](https://docs.docker.com/reference/api/engine/v1.45/#tag/Container/operation/ContainerList)
* [Golang SDK function](https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerList)
* One existing call site of relevant Golang function:
https://github.com/kreuzwerker/terraform-provider-docker/blob/54685a5fb9b375961620d2c555c7f3925327112c/internal/provider/resource_docker_container_funcs.go#L917

I couldn't find any issues that seemed to match the need for listing containers for enumeration, which I'm surprised by. Sorry if I missed a relevant existing issue. 

I am a Golang noob, and I've never written a Terraform Provider before, but this might be a good place for me to cut my teeth. Wanted to flag it for consideration to see if this has already been discussed as a Won't Do or something. 

Thanks for your consideration

PR #644, opened 2024-08-24
feat: add new data source for docker_containers

KetchupBomb:
Addresses https://github.com/kreuzwerker/terraform-provider-docker/issues/642 by adding a new data source -- `docker_containers` -- for listing containers.

Right now, the data source has no schema beyond what is computed. The primary attribute, `containers`, is currently a list of strings. This is a naive implementation that can be improved, but is still quite compatible with Terraform through `jsondecode()`:

```hcl
> jsondecode(data.docker_containers.this.containers[0]).Id
"037d3b70a777ad7f66ae60534545fed5d69c784cdf685810370ed4f2cfed9020"
```

[The Docker Engine API endpoint](https://docs.docker.com/reference/api/engine/v1.45/#tag/Container/operation/ContainerList) exposes more parameters that can be incorporated into this data source, but I wanted to put this up for initial consideration and guidance for me as a Golang noob, and first time working with Terraform Provider development. 

- The Docker Engine API returns values without ordering, so as of right now, repeated `terraform plan` or `terraform apply` commands will likely report changes to the underlying data source. 
- I am unsure of how to write adequate testing for such a simple case. Even then, I noticed that purposefully writing incorrect test cases would not cause `make test` or `go test` to fail. Advice on making sure I am writing tests correctly would be appreciated.


Junkern:
Thanks for the PR and sorry for the delay. I already added this functionality as part of #893 
Feel free to use and and report any bugs!

PR #646, opened 2024-08-29
feat: Support registries that return empty auth scope

legioner0:
I've found private docker registry service on hosting provider, that returns "www-authenticate" header without scope. Here is fix for that case.


Junkern:
@legioner0 sorry, I saw your update too late and now you have conflicts again... Do you mind rebasing one last time and then pinging me so that we can get this merged? Thank you!

Junkern:
Thank you!

Issue #534, opened 2023-03-24
failed to read downloaded context: failed to load cache key: invalid response status 403


atali:
Hello,

I have a weird behaviour, trying to build a simple docker image but it failed with the following error :

```
Terraform will perform the following actions:

  # docker_image.my-image will be created
  + resource "docker_image" "my-image" {
      + id          = (known after apply)
      + image_id    = (known after apply)
      + name        = "my-image:latest"
      + repo_digest = (known after apply)

      + build {
          + cache_from   = []
          + context      = "/EruUkohXdib6p2nB0KVIFDv54yk/HpGfE2XOJY4zO8Sxa1eXbxzHPjk"
          + dockerfile   = "/EruUkohXdib6p2nB0KVIFDv54yk/HpGfE2XOJY4zO8Sxa1eXbxzHPjk/mydocker/Dockerfile"
          + extra_hosts  = []
          + remove       = true
          + security_opt = []
          + tag          = []
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
docker_image.my-image: Creating...
│ 
│ Error: failed to read downloaded context: failed to load cache key: invalid response status 403
│ 
│ 
│ 
│   with docker_image.my-image,
│   on main.tf line 13, in resource "docker_image" "my-image":
│   13: resource "docker_image" "my-image" {
│ 
╵
```

Below the terraform:

```
terraform {
  required_providers {
    docker = {
      source  = "kreuzwerker/docker"
      version = "~> 3.0.2"
    }
  }
}

provider "docker" {
}

resource "docker_image" "my-image" {
  name = "my-image:latest"

  build {

    context = "${path.cwd}"
    dockerfile = "${path.cwd}/mydocker/Dockerfile"
  }
}
```

This is the content of the folder /EruUkohXdib6p2nB0KVIFDv54yk/HpGfE2XOJY4zO8Sxa1eXbxzHPjk:

![image](https://user-images.githubusercontent.com/369628/227510556-3832d0df-f062-4620-a42d-021b38edddbe.png)

when I cd to  /EruUkohXdib6p2nB0KVIFDv54yk/HpGfE2XOJY4zO8Sxa1eXbxzHPjk and build with docker , it works

```
docker build -t my-image:latest ./mydocker
```

My dockerfile :
```
# syntax=docker/dockerfile:1
FROM busybox

RUN echo "SUCCESS"
```
Does someone has an idea ?

atali:
[...]

PR #665, opened 2024-12-28
Document Requirement To Disable containerd For Image Build

tparikka:
Documents the need to disable containerd in order to support build of image in Docker, as described at #534.


Junkern:

PR #677, opened 2025-04-11
fix: Set correct default network driver and fix a test

Junkern:
Just trying to fix tests and get everything running again. Default docker networking has changed with newer docker server version. I am also thinking of adding tests to test against different docker server versions, e.g. via a matrix as we do with terraform


Junkern:
@t3r please review :)

t3r:
merged, thanks!

PR #762, opened 2025-07-26
feat: Add CDI device support

jdon:
This PR allows users to use the CDI interface to pass through GPUs to their containers instead of using the existing `gpus  = "all"`.

This is needed as operating systems like [nixos](https://discourse.nixos.org/t/virtualisation-docker-enablenvidia-vs-hardware-nvidia-container-toolkit-enable/51370) are removing the support for `gpus all`.

### Changes:
  - Adds CDI (Container Device Interface) device support via a new device_requests field
  - Added ConflictsWith validation between gpus and device_requests
  - Updated documentation
  - Added tests for:
    - CDI device
    - GPU
    - Conflict validation
  - Added check in tests to skip them if a NVIDIA GPU is not found.

### DIFF
--- a/docs/data-sources/image.md
+++ b/docs/data-sources/image.md
@@ -45,5 +45,3 @@ data "docker_image" "tag_and_digest" {
 
 - `id` (String) The ID of this resource.
 - `repo_digest` (String) The image sha256 digest in the form of `repo[:tag]@sha256:<hash>`. It may be empty in the edge case where the local image was pulled from a repo, tagged locally, and then referred to in the data source by that local name/tag.
-
-
--- a/docs/data-sources/logs.md
+++ b/docs/data-sources/logs.md
@@ -36,5 +36,3 @@ description: |-
 
 - `id` (String) The ID of this resource.
 - `logs_list_string` (List of String) List of container logs, each element is a line.
-
-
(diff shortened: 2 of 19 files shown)


Junkern:
Thanks for your PR! I spent some minutes looking into the codebase to understand the context and your request.
I might be wrong, but what you are trying to achieve is already possible/implemented?

I think the documentation around the `gpu` parameter is wrong: 
* `Currently, only the value `all` is supported.`  => this is not correct - we pass the value of the `gpus` atttribute to the docker API exactly as the Docker CLI does.

But I suppose you tried access an nVidia GPU (https://docs.docker.com/reference/cli/docker/container/run/#gpus) and it didn't work?

Then we might be missing a small detail in the implementation

jdon:
Hey @Junkern, 

Thanks for the ping here, I dropped the ball on this, I've just been using a fork locally for my needs :joy:.

It would be great to get this merged! I've just fixed up the conflicts. Please let me know what changes you would like and I'm more than happy to make them.

> (In addition to this PR I will fix the gpus implementation)

What are your plans for this? Are you planning on removing `gpus` and pushing everyone down the device requests route?


(Also feel free to push commits to this branch if that's quicker / easier for you!)

PR #791, opened 2025-09-18
feat: Implement docker init in docker service

mx-moth:
This PR is a duplicate of #760, but with your requested updates. I found myself wanting to use the `--init` flag in a service and found the stale PR.

I am unfamiliar with go generally, so this may not be entirely correct. It does compile! But I am unsure how to test it.


mx-moth:
I've made changes and it now compiles locally for me using `make build`. Running `make test` fails as I have added a required parameter to one of the functions being tested. I attempted to update the test code as well but this is beyond my non-existent go skills. The function being tested needs a DockerClient instance but I don't know how to make a client for test purposes.

Junkern:
@mx-moth to save time, can you instruct copilot to fix the failing test? It will also know how to instantiate the needed DockerClient instance :)

Junkern:
Thanks for the contribution! I did now implement it via #949  Will be released within the next week :)

PR #937, opened 2026-05-14
feat: Add no_compose build tag for non-CGO cross-compile

pose:
## Summary

Adds an opt-in `no_compose` build tag that excludes `resource_docker_compose.go` (and its transitive `docker/compose/v2` import) from the compiled provider. Consumers building this provider via static cross-compile with `CGO_ENABLED=0` can use the tag to keep the rest of the provider buildable.

## Motivation

`resource_docker_compose.go` imports `github.com/docker/compose/v2/pkg/compose`, which always pulls in `pkg/watch`, which on darwin imports `github.com/fsnotify/fsevents`. `fsnotify/fsevents` requires CGO at build time — its API is split across files tagged `//go:build darwin` and `//go:build darwin && cgo`, and the symbols defined in the latter are referenced unconditionally from the former. Building the provider with `GOOS=darwin CGO_ENABLED=0` therefore fails with:

```
fsevents.go:147: undefined: GetDeviceUUID
fsevents.go:148: es.start undefined
fsevents.go:161: undefined: flush
fsevents.go:167: undefined: stop
```

For most consumers (including this provider's own release pipeline, which builds natively per platform with CGO=1), this is invisible. But downstream projects that cross-compile the provider from Linux for darwin distribution need a way to opt out of the compose code path, because no version of `fsnotify/fsevents` supports `CGO_ENABLED=0` and the import is currently unconditional.

Concrete downstream impact: pulumi/pulumi-docker bridges this provider and ships static binaries cross-compiled from Linux for darwin-amd64 and darwin-arm64. Their master CI has been red since v4.3.0 because of this issue ([pulumi/pulumi-docker#1698](https://github.com/pulumi/pulumi-docker/issues/1698)). They are shipping a temporary patch that deletes the resource entirely; this PR would let them ship the full provider minus just the compose resource when the tag is set.

There is also a related upstream request in [docker/compose#13448](https://github.com/docker/compose/issues/13448) asking compose itself to make watch opt-in, which would solve the same problem at a different layer.

## Implementation

- `resource_docker_compose.go`: add `//go:build !no_compose`.
- Split the resource registration in `framework_provider.go` into two files, gated by the same tag:
  - `compose_resources_enabled.go` (`//go:build !no_compose`) returns `[NewDockerComposeResource]`.
  - `compose_resources_disabled.go` (`//go:build no_compose`) returns `nil`.
- `framework_provider.go`'s `Resources()` calls the build-tag-selected `composeResources()` helper instead of inlining the slice.

Default behavior is unchanged: a plain `go build ./...` includes `docker_compose`. The tag is purely opt-in.

## Verification

```bash
# Default build — compose included
go build ./...
# succeeds

# Opt-out, cross-compile darwin with CGO=0 — compose excluded
GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go build -tags no_compose ./...
# succeeds

# Sanity: same cross-compile without the tag
GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go build ./...
# fails with fsevents undefined symbols (current master behavior)
```

## Notes

- I left the existing `docker/compose/v2` requires in `go.mod` untouched, since they're still used by the default build. A future PR could move them under a build-constrained require if desired, but that's optional.
- Happy to adjust the tag name (`no_compose` vs `nocompose` vs `cgo` opt-in) or the file naming conventions based on maintainer preference.


pose:
Considering alternative directions 🙏  Thanks

pose:
Alternatives considered and I don't think there's no easier workaround. I would appreciate a review now. Thanks! 🙏

Junkern:
Really interesting PR! I will take a closer look in the upcoming days. In the best case this is a no-brainer for merging and will be released with the next release

pose:
Thanks @Junkern!

Junkern:
@pose  I also took a look at the linked issue inside https://github.com/docker/compose/issues/13448 and I think the cleanest approach would be to bump our `docker/compose` dependency to at least `5.1`. That way we do not introduce our own buildtag, but you can rely on the one from `docker/compose`. I will open a feature branch today - maybe you can test it out?

Junkern:
I will merge this, as the bumping of dependencies will trigger some bigger reworks (which I want to avoid for now)

pose:
Thanks!

Junkern:
@pose Are there any other `pulumi-docker` issues that would require changes in this provider? Happy to help!

pose:
@Junkern nothing for now, but thanks for the offer! 🙏

PR #959, opened 2026-06-24
fix: surface failed task logs on service converge timeout

### LINKED ISSUE #624: Support for error message when service didn't converge
### Community Note

### Description

When specifying a converge config on a `docker_service` resource and the initial deployment of that service fails, e.g. because the container started as part of the service exits right away because of misconfiguration, then there is no way to see the error message, because the provider will delete the service and the failed tasks with it after the converge time is over.

### New or Affected Resource(s)

* docker_service

nivednarayan:
Fixes #624

## Problem
When a `docker_service` fails to converge (e.g. the container exits 
immediately due to misconfiguration), the provider deletes the service 
and its tasks before returning the error. This leaves the user with no 
information about why the service failed:

Error: Service with ID (abc123) did not converge after 30s

## Solution
Before deleting the service, fetch the last 10 log lines from any 
failed tasks and surface them in the error message:

Error: Service with ID (abc123) did not converge after 30s

Last task logs:
failing on purpose
failing on purpose

## Testing
Tested locally with a deliberately broken service (container exits 
immediately with exit code 1) and confirmed logs are surfaced in the 
error. Also tested with a healthy service to confirm no issue on 
the happy path.

### DIFF
--- a/internal/provider/resource_docker_service_funcs.go
+++ b/internal/provider/resource_docker_service_funcs.go
@@ -1,6 +1,7 @@
 package provider
 
 import (
+	"bufio"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


Junkern:
Thanks for the PR @nivednarayan Will be part of the next `4.6.0` release :)

